PR #2, opened 2018-10-18
Add separate builder for journalbeat

kvch:
New separate `Dockerfile` is added to build `journalbeat`. I introduced a new builder, because the vendored lib which accesses journals requires `libsystemd-dev` package. I would like to keep general builder images minimal. As right now `journalbeat` is the only one which requires this package, I do not want to install it in all containers.

The tag of the new image is suffixed with `-journalbeat`. I made builder tag suffixes configurable in `mage` of Beats. I will open the PR in Beats after this PR is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ build:
 	@make -C go1.11 $@
 	@make -C go1.11 -f Makefile.debian7 $@
 	@make -C fpm $@
+	@make -C go1.10-journalbeat $@
+	@make -C go1.10-journalbeat -f Makefile.debian7 $@
 
 # Requires login at https://docker.elastic.co:7000/.
 push:
--- a/go1.10-journalbeat/Makefile
+++ b/go1.10-journalbeat/Makefile
@@ -1,0 +1,10 @@
+IMAGES := base main
+
+build:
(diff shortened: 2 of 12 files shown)


kvch:
In case of `packetbeat` `libpcap` is compiled on the target system. It is true that both `libsystemd-dev` and `libpcap` are Beat-specific dependencies. However, I think it is better to install `libsystemd-dev` in the container in advance, because `apt` takes care of all dependencies of this lib. If I run it using `mage` I need to use `dpkg`, because `apt` is not available. I need to download each dependency of the packages "manually", because `dpkg` cannot fetch/install them automatically. I think it is simpler to have a separate builder in this case.

tsg:
@kvch It's just that this creates a bit of duplication between go1.10 and go1.10-functionbeat. We had separate images in the previous system and that was a bit messy to maintain. @andrewkroh cleaned it up and reduced everything to one image so he might not be thrilled about this change :)

How about we just always install `libsystemd-dev`, even if not needed by the other Beats? I'm not really worried about the image being slightly larger, are there other concerns?

kvch:
No, nothing. I am adding the package in separate PR.

ph:
@tsg look like a good compromise

PR #3, opened 2018-10-19
Add `libsystemd-dev` to packages of base image

kvch:


andrewkroh:
These images do not include any Beat specific dependencies. That is entirely handled by the thing that is using these images. When using these images you control what is invoked inside of the container. So if you are building a Go binary with a Makefile target, that target could have a dependency on another target that installs the necessary dependencies (e.g. `apt-get update && apt-get -y install libsystemd-dev`).

In some cases the same image contains cross-compilers for multiple architectures (like mips/mips64/mips64le OR amd64/i386). For this image you would need to install the same package for [multiple architectures](https://wiki.debian.org/Multiarch/HOWTO) in order to have both `/lib/mips-linux-gnu/libsystemd.so` and `/lib/mips64el-linux-gnuabi64/libsystemd.so`, for example. So if you continue to go this route of pre-installing libsystemd-dev then you need to take this into account.

My recommendation is to first try installing libsystemd-dev in the running container prior to executing the build. Can you see if that works for this?

kvch:
I tried following the pattern in `packetbeat` if that is what you mean. However, I gave up when I had to collect the 5th "layer" of dependencies manually. It felt like a waste of time to collect all packages including the one I would like to use, then its dependencies and the dependencies of the dependencies when I have access to `apt` during image build time. (Debian 7 requires different packages than the latest one. So it means collecting even more packages and their dependencies.)

I don't think it's worth the effort of searching for all packages manually and installing them in case of an experimental Beat. Also, when I had to download packages from `"security"`, I even felt more uneasy, because https://www.debian.org/distrib/packages warned me in big red letters not to download such packages from the site, but rather use `apt`.

I think in case of Packetbeat is completely acceptable that it compiles `libpcap` on the builder. Because it is part of the build process. But installing required packages is part of setting up the environment which I think needs to be done in a `Dockerfile`. I agree, adding this to the base image is not the best solution. I still think that the best way is to have a separate builder image for which is based on top of the builders you have created. But again, I am willing to compromise.

PR #4, opened 2018-10-23
Add new Debian builder && rm mage from images

kvch:


andrewkroh:
The top-level Makefile should be updated to execute the builds for debian8 images. https://github.com/elastic/golang-crossbuild/blob/master/Makefile

PR #13, opened 2019-01-31
update to FPM 1.11.0

ph:
Previous version of FPM did not pin to a specific version of
childprocess.

See for details https://github.com/jordansissel/fpm/issues/1592


ph:
docker images pushed.

PR #16, opened 2019-03-22
Use archive.debian.org for apt sources on debian 7

andrewkroh:
The debian 7 (wheezy) mirrors were archived since it's EOL.
I don't think we statically link with anything from wheezy we just use it because it
has an early version of glibc so we get decent OS compatibility. But we should
still consider moving to a new version.

https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

I added templating so I could make the inclusion of the sources.list file conditional. This could be expanded in the future to reduce some duplication or do other cleanup.


andrewkroh:
I pushed updated images for 1.10.8, 1.11.5, 1.11.6. 

For 1.11.5 I used this commit, but I changed the Go version and download hash locally.

PR #19, opened 2019-04-10
Currently deal with debian7 EOL by allow to use security packages.

ph:


ph:
1.10.8 1.11.5 and 1.11.6 were pushed.

PR #21, opened 2019-04-12
Fix the fpm targets

ph:
Just build fpm out of the target loop, so we don't have to check if the `Makefile.debian*` exists, since we rarely change fpm we should build it often.


andrewkroh:
I didn't notice this b/c I hacked the target list to be just go1.12 for testing.

kvch:
Thank you for catching my error.

PR #39, opened 2020-01-27
Add dependencies for x-pack/auditbeat

mikemadden42:


go1.13/arm/Dockerfile.tmpl
```diff
@@ -10,6 +10,9 @@ RUN \
         linux-libc-dev-arm64-cross \
         linux-libc-dev-armel-cross \
         linux-libc-dev-armhf-cross \
+        librpm-dev:arm64 \
```
andrewkroh: Has this actually been tested? I don't think that will work without first doing an `--add-architecture` call.

andrewkroh: And this may result in conflicts that require different images for different architectures.

mikemadden42: Nice catch.  I'll revise the PR to add the arch as well.

```
E: Unable to locate package librpm-dev:arm64
E: Unable to locate package librpm-dev:armhf
E: Unable to locate package librpm-dev:armel
```

andrewkroh: BTW the Debian [cross-compile guide](https://wiki.debian.org/CrossCompiling#Set_up_a_chroot) recommends using a chroot to deal with those Multi-arch conflicts. But since we're in Docker we can just use a different image for each arch to avoid conflicts.

mikemadden42: Can you clarify the conflicts.  Should it be ok as long as we specify `pkg:arch` like `librpm-dev:arm64`?

kuisathaverat:
superseded by https://github.com/elastic/golang-crossbuild/pull/58

PR #48, opened 2020-04-27
Add Debian 9 images

jsoriano:


fatmcgav:
I think `9` is the default `DEBIAN_VERSION`, based on: https://github.com/elastic/golang-crossbuild/blob/126ff6fc22fb66ebaaeb194940e066b5ed8551de/go1.13/Makefile.common#L6

jsoriano:
Oh, you are right.

Issue #23, opened 2019-04-25
golang-cross build images should contain all needed build components for beats


mikemadden42:
We ran into an [issue](https://internal-ci.elastic.co/job/elastic+release-manager+release-staging/129/console) where we could not install `libsystemd-dev:i386` which is needed to cross compile journalbeat.  It would be nice if this package & its requirements were included in the golang-crossbuild images.

```
08:12:56 E: Unable to correct problems, you have held broken packages.
08:12:56 Error: running "apt-get install -y --no-install-recommends libsystemd-dev:i386" failed with exit code 100
08:12:56 Error: failed building for linux/386: exit status 100
08:12:56 failed building for linux/386: exit status 100
08:12:56 E: Unable to correct problems, you have held broken packages.
08:12:56 Error: running "apt-get install -y --no-install-recommends libsystemd-dev:i386" failed with exit code 100
08:12:56 Error: failed building for linux/386: exit status 100
08:12:56 failed building for linux/386: exit status 100
```

andrewkroh:
I agree. I was coming to open the same issue and found this.

Originally I opted for having this project be generic and only have the core dependencies necessary for running a cross-compiler. But this restriction causes too many problems when the upstream package repos change / go offline / or are retired so embedding the Debian packages necessary for Beats will help and will make builds more reproducible.

mikemadden42:
Initial PR for `x-pack/auditbeat` updates is available (https://github.com/elastic/golang-crossbuild/pull/39).

v1v:
I'll close this issue since it was done but the `zube` bot decided to reopen it :(

PR #52, opened 2020-06-03
Add pcap, rpm, systemd packages to main image.

mikemadden42:


mikemadden42:
Related issue:
https://github.com/elastic/golang-crossbuild/issues/23

go1.10/main/Dockerfile.tmpl
```diff
@@ -12,6 +12,12 @@ RUN \
         gcc-multilib \
         libc6-dev \
         libc6-dev-i386 \
+        libpcap-dev \
```
andrewkroh, line 15: Packetbeat builds libpcap from source due to some issues we had to resolve:

- https://github.com/elastic/beats/blob/abae05768c4c445bddffd6fff520a62ccca2e126/packetbeat/magefile.go#L260-L262
- https://github.com/elastic/beats/pull/4217

I think you will have to do the same thing in the image if you fully want to make the image fully self-contained. Did you already test building Packetbeat with this image using the OS provided libpcap?

Addressing libpcap could be left until the end. It's not really a blocker for the issue you are trying to resolve. Having it in the image will speed up the builds, but we don't utilize any `apt` commands from Packetbeat's build to get libpcap built.

mikemadden42: I removed the `libpcap` libraries for now.  We can reassess it at the end when all the other packages are addressed.

mikemadden42:
I'm running into an issue building the image after adding the `librpm` and `libsystemd-dev` packages.

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 librpm-dev : Depends: librpm3 (= 4.11.3-1.1) but it is not going to be installed
              Depends: librpmio3 (= 4.11.3-1.1) but it is not going to be installed
              Depends: librpmbuild3 (= 4.11.3-1.1) but it is not going to be installed
              Depends: librpmsign1 (= 4.11.3-1.1) but it is not going to be installed
              Conflicts: librpm-dev:i386 but 4.11.3-1.1 is to be installed
 librpm-dev:i386 : Depends: libdb-dev:i386 but it is not going to be installed
                   Conflicts: librpm-dev but 4.11.3-1.1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

mikemadden42:
If I remove `librpm-dev:i386`, the build succeeds.

andrewkroh:
You're running into a mutliarch conflicts. I think the only way to solve this is to have separate images for `amd64` and `386`. You're going to hit those issues with the arm build image too.

mikemadden42:
I'm going to close this PR. I've started with a better approach in this PR:
https://github.com/elastic/golang-crossbuild/pull/53

PR #54, opened 2020-06-12
[CI]: CI pipeline

v1v:
### What

Enable CI pipeline to build and push the docker images.

The push event works for:
- everything (branches, tags, PRs) and will push them in https://docker.elastic.co:7000/observability-ci/golang-crossbuild:<VERSION><SUFFIX> 
- on master based to the default location set in the Makefile.

### Tasks

- [x] Create JJBB definition (see https://github.com/elastic/beats/pull/19162)

### Test

- `REPOSITORY=docker.elastic.co/observability-ci make build`

```
$ docker images | grep observability-ci            
docker.elastic.co/observability-ci/fpm                      1.11.0                 1fbf2b9b5555        2 minutes ago       416MB
docker.elastic.co/observability-ci/golang-crossbuild        1.14.4-main-debian8    51fbb901cd8a        2 minutes ago       1.39GB
docker.elastic.co/observability-ci/golang-crossbuild        1.14.4-main-debian7    3c5321b6f233        2 minutes ago       1.01GB
docker.elastic.co/observability-ci/golang-crossbuild        1.14.4-s390x           8064f3af14bd        2 minutes ago       807MB
docker.elastic.co/observability-ci/golang-crossbuild        1.14.4-ppc             57b58c057ff9        2 minutes ago       994MB
...
```

- `REPOSITORY=docker.elastic.co/observability-ci make push`

```
echo ">> Pushing docker.elastic.co/observability-ci/golang-crossbuild:1.10.8-base"
>> Pushing docker.elastic.co/observability-ci/golang-crossbuild:1.10.8-base
The push refers to repository [docker.elastic.co/observability-ci/golang-crossbuild]
772921e6e3a7: Preparing 
041e855fb031: Preparing 
e142d45a5747: Preparing 
...
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,14 +8,13 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     NOTIFY_TO = credentials('notify-to')
     PIPELINE_LOG_LEVEL = 'INFO'
-    PATH = "${env.PATH}:${env.WORKSPACE}/bin"
     HOME = "${env.WORKSPACE}"
     DOCKER_REGISTRY_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.REGISTRY}/observability-ci"
   }
   options {
-    timeout(time: 1, unit: 'HOURS')
+    timeout(time: 2, unit: 'HOURS')
     buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20', daysToKeepStr: '30'))
     timestamps()
     ansiColor('xterm')
@@ -49,12 +48,10 @@ pipeline {
     stage('Staging') {
       steps {
         withGithubNotify(context: 'Staging') {
-          deleteDir()
-          unstash 'source'
           dir(BASE_DIR){
             dockerLogin(secret: "${DOCKER_REGISTRY_SECRET}", registry: "${REGISTRY}")
-            sh(label: "push docker image to ${env.STAGING_IMAGE}/${env.REPO_NAME}",
-               script: "REPOSITORY=${env.STAGING_IMAGE}/${env.REPO_NAME} VERSION==${env.BRANCH_NAME} make push")
+            sh(label: "push docker image to ${env.STAGING_IMAGE}/${env.REPO}",
+               script: "REPOSITORY=${env.STAGING_IMAGE}/${env.REPO} VERSION=${env.BRANCH_NAME} make push")
           }
         }
       }
@@ -69,8 +66,6 @@ pipeline {
         stage('Publish') {
           steps {
             withGithubNotify(context: 'Publish') {
-              deleteDir()
-              unstash 'source'
               dir(BASE_DIR){
                 dockerLogin(secret: "${DOCKER_REGISTRY_SECRET}", registry: "${REGISTRY}")
                 sh 'make push'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent { label 'linux && immutable' }
+  agent { label 'ubuntu && immutable' }
   environment {
     REPO = 'golang-crossbuild'
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,17 +49,18 @@ pipeline {
       }
     }
     stage('Staging') {
+      environment {
+        REPOSITORY = "${env.STAGING_IMAGE}"
+      }
       steps {
         withGithubNotify(context: 'Staging') {
           withGoEnv(){
             dir(BASE_DIR){
               dockerLogin(secret: "${DOCKER_REGISTRY_SECRET}", registry: "${REGISTRY}")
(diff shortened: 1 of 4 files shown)


v1v:
I need your review:
- @urso , @andrewkroh , @elastic/observablt-robots 

I'm afraid I'm not an admin of this particular repo therefore I cannot assign reviewers.

Jenkinsfile
```diff
@@ -0,0 +1,88 @@
...
+  }
+  triggers {
+    issueCommentTrigger('(?i).*jenkins\\W+run\\W+(?:the\\W+)?tests(?:\\W+please)?.*')
+  }
+  stages {
+    stage('Checkout') {
+      steps {
+        deleteDir()
+        gitCheckout(basedir: BASE_DIR)
+        stash allowEmpty: true, name: 'source', useDefaultExcludes: false
```
cachedout, line 34: Nit: should `allowEmpty` really be set to `true`? Seems like subsequent stages are sure to fail if this stash is empty.

Jenkinsfile
```diff
@@ -0,0 +1,88 @@
...
+        }
+      }
+    }
+  }
+  post {
+    always {
+      notifyBuildResult()
+    }
+  }
+}
```
cachedout, line 88:
```suggestion
}

```

Super-nit: trailing newline :)

Jenkinsfile
v1v:
```suggestion
        stash name: 'source', useDefaultExcludes: false
```

Makefile.common
```diff
@@ -1,4 +1,4 @@
-REPOSITORY    := docker.elastic.co/beats-dev
+REPOSITORY    ?= docker.elastic.co/beats-dev
```
v1v, line 1: This is the only change in the code to be able to set the repository where the docker images can be pushed, by default as used to be, but as we will be using the `staging` stage, then this will help to generate docker images on a PR basis which can be consumed for testing purposes.

andrewkroh:
> I'm afraid I'm not an admin of this particular repo therefore I cannot assign reviewers.

This repo is tied in with membership in https://github.com/orgs/elastic/teams/beats-contributors if you want to join.

andrewkroh:
> on tag based to the default location set in the Makefile.

This implies that this repo will need to start doing some tagging. As of now we've been pushing new images whenever master changed. And we really only changed master when new Go releases were available. I'm not against requiring tagging, but we should have a plan for it if we go this route (when do we tag, what do the tags mean, how do we convey the tag in the image metadata).

v1v:
@andrewkroh , thanks for the details, I'm not really familiar with the release process for this project, I was just thinking the tag could be matching the go version, but there are certain questions as you stated that they are required to be addressed beforehand. 

What do we do from here?

kuisathaverat:
as Victor sad, the most logical tag is the Go version. Using tags adds the possibility to compile with any images generated for a previous Go version in case we need it for some test, or we want to compile with several Go versions at some point.

andrewkroh:
To be clear I was talking about git tagging, which is what I assume is meant in "on tag based to the default location". Any build of the repo produces images for multiple Go versions. Those images have docker tags containing the Go version and the cross-compiler architecture (see [build tags](https://github.com/elastic/golang-crossbuild#build-tags)).

Perhaps the path forward is to not introduce git tagging into the image push process at this time. And setup the job to push changes from master to the default location in the Makefile.

I think we can get the same ability to reproduce old builds if we add an additional docker tag (without git tags) that contains a time component (e.g. 1.10.8-main-20191124). Then if I need to exactly reproduce some old Beats build for whatever reason, we can use the time tagged docker images from the time of the original build. The images already contain docker labels following http://label-schema.org/rc1/ that contain this repo's commit ID in case we need to know the exact Dockerfiles that went into producing the images.

If we do want to use git tags to trigger the image pushes then we should
- push an additional docker tag that incorporates the git tag (e.g. push both `1.14.1-main-debian7` and `1.14.1-main-debian7-vXX`)
- add the git tag to the image labels (org.label-schema.version = $git_tag)

v1v:
To keep it simple, I'll remove the `on tag based` event to be `master based`. Then we can start a conversation in another thread to discuss how the release process looks like and what are the things, if any, that we could change.

v1v:
https://github.com/elastic/golang-crossbuild/pull/54/commits/c5cc23429827378a199e9989d1955d1bb6ebd9be is now the new behaviour. 

_NOTE_: if we merge this particular PR then the automation to push those docker images will happen immediately.

v1v:
Alright, we got all the required webhooks and service accounts in place to run in the CI!

As soon as the build gets green then I'll merge it

v1v:
For the record https://github.com/elastic/golang-crossbuild/pull/54/commits/ead3be24997023206d6f0c35ae25df389013603a removed the issues with the push.* which it's not required anymore at all (for furtherh details please refers to the infra docs).

Already tested locally:

```bash
$ docker tag docker.elastic.co/observability-ci/fpm:1.11.0 docker.elastic.co/observability-ci/fpm:test                                                                       
$ docker push docker.elastic.co/observability-ci/fpm:test                                                                   
The push refers to repository [docker.elastic.co/observability-ci/fpm]
aa5033bd89aa: Pushed 
dc29c4af552e: Pushed 
0f7b4d7b24a5: Pushed 
7deb88e599ba: Pushed 
8354d5896557: Mounted from observability-ci/golang-crossbuild 
test: digest: sha256:1c8721bf3d8095878b02855a8c3d90a6c6d380de49c2483929e68323df487687 size: 1368
```

PR #53, opened 2020-06-11
Make golangcross-build images standalone

mikemadden42:
- Rename main to amd64

PR #58, opened 2020-07-07
Add dependencies for x-pack/auditbeat and Journalbeat

kuisathaverat:
it installs cross-compile dependencies into the cross-build Docker images, finally, the trick is to add all the cross-packages manually, it seems it does not follow the dependencies.

This PR superseded https://github.com/elastic/golang-crossbuild/pull/39


go1.13/arm/Dockerfile.tmpl
```diff
@@ -3,13 +3,49 @@ ARG VERSION
...
+        libc-dev:arm64 \
+        libpopt-dev:arm64 \
+        linux-libc-dev:arm64 \
+        libxml2-dev:arm64 \
+        libxml2:arm64 \
+        libicu-dev:arm64 \
+        libicu57:arm64 \
+        icu-devtools:arm64 \
+        libsystemd-dev:arm64 \
+#        librpm-dev:armhf \
```
kuisathaverat, line 29: we cannot install the three ARM variants, the packages conflict with each other. I have to take a look at https://wiki.debian.org/CrossCompiling to see if there is a way to make it. An alternative is to have an image for each ARM variant

go1.13/main/Dockerfile.tmpl
```diff
@@ -17,6 +17,7 @@ RUN \
         mingw-w64 \
         mingw-w64-tools \
         patch \
+        librpm-dev:i386 \
```
urso, line 20: @kvch I think we are missing journalbeat dependencies here. What is required for amd64 and i386 builds (I think i386 requires us to jump through some extra hoops).

Is the `main` image used for building i386 and amd64 builds? Maybe we should provide separate images. Especially the systemd dependency can be a pain.

kuisathaverat: I guess so, but I could find a place where are defined the packages needed for any beat or the place where they are installed when you run the `mage package`

kuisathaverat:
BTW this generates images for the Go version 1.13.12 (among others), we use 1.13.10 on Beats, so or we bump the version in Beats or we generate a version 1.13.10 here

go1.13/main/Dockerfile.tmpl
kuisathaverat: there is a PR that create a new amd64 Docker image https://github.com/elastic/golang-crossbuild/pull/53, I will make a review of it with the changes in here

kuisathaverat:
@urso I have changed the Go version to 1.13.10 the same Beats use, to use the generated images in the package job

urso:
+1 for providing 1.13.10.

We still try to update to go1.14.4: https://github.com/elastic/beats/pull/18829

go1.13/main/Dockerfile.tmpl
kvch: @urso Mage installs the headers Journalbeat needs before running the default `GolangCrossBuild` target. Previously it was not added to the common Docker image to avoid installing unnecessary packages for different Beats. It is similar to Packetbeat, there the unique dependencies of the Beat are installed before crossbuild. I think it has been a good practice. I would not keep the common images as minimal as possible.

kvch: I am not sure I follow why this PR is needed. Journalbeat installs its dependencies before running GolangCrossbuild. Once it is merged, we need to follow-up in the magefile of the Beat?

kvch: I had a chat with @urso offline about the goal of this PR. You can find the deps of Journalbeat by platform here: https://github.com/elastic/beats/blob/master/journalbeat/magefile.go#L121-#L133

Issue #60, opened 2020-07-08
Docker images fot 1.13.10 are not pushed correctly


kuisathaverat:
The push commands for the 1.13.10 Docker images are failing

```
9:27:27  >> Pushing docker.elastic.co/beats-dev/golang-crossbuild:1.13.10-base-debian7
19:27:27  The push refers to repository [docker.elastic.co/beats-dev/golang-crossbuild]
19:27:27  tag does not exist: docker.elastic.co/beats-dev/golang-crossbuild:1.13.10-base-debian7
19:27:27  ..//../Makefile.common:9: recipe for target 'push' failed
19:27:27  make[2]: *** [push] Error 1
19:27:27  make[2]: Leaving directory '/var/lib/jenkins/workspace/ats_golang-crossbuild-mbp_master/src/github.com/elastic/golang-crossbuild/go1.13/base'
19:27:27  Makefile.debian7:12: recipe for target 'push' failed
19:27:27  make[1]: *** [push] Error 1
```

cc @elastic/observablt-robots

kuisathaverat:
There is a previous error related to the checksum of Go

```
18:45:50  Step 7/18 : RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz 	&& echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - 	&& tar -C /usr/local -xzf golang.tar.gz 	&& rm golang.tar.gz
18:45:50   ---> Running in 0174f1e03558
18:45:52  sha256sum: WARNING: 1 computed checksum did NOT match
18:45:52  golang.tar.gz: FAILED
18:45:53  The command '/bin/sh -c curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz 	&& echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - 	&& tar -C /usr/local -xzf golang.tar.gz 	&& rm golang.tar.gz' returned a non-zero code: 1
18:45:53  ../Makefile.common:13: recipe for target 'build' failed
18:45:53  make[2]: *** [build] Error 1
18:45:53  make[2]: Leaving directory '/var/lib/jenkins/workspace/ats_golang-crossbuild-mbp_master/src/github.com/elastic/golang-crossbuild/go1.13/base'
18:45:53  Makefile:4: recipe for target 'build' failed
18:45:53  make[1]: *** [build] Error 1
```

PR #61, opened 2020-07-08
fix: set the proper sha256

kuisathaverat:
I did not change the sha256 of the Go version

close https://github.com/elastic/golang-crossbuild/issues/60


go1.13/base/Dockerfile.tmpl
```diff
@@ -23,7 +23,7 @@ RUN \
 
 ARG GOLANG_VERSION=1.13.10
 ARG GOLANG_DOWNLOAD_URL=https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz
-ARG GOLANG_DOWNLOAD_SHA256=9cacc6653563771b458c13056265aa0c21b8a23ca9408278484e4efde4160618
+ARG GOLANG_DOWNLOAD_SHA256=8a4cbc9f2b95d114c38f6cbe94a45372d48c604b707db2057c787398dfbf8e7f
```
mdelapenya, line 26: Just for future reference: where does this sha come from?

kuisathaverat: https://golang.org/dl/

Screenshot 2020-07-08 at 20 20 27

PR #62, opened 2020-07-20
Fix/go14.4.6

kuisathaverat:
* bump go version to 1.14.6
* split ARM arch
* split x86 arch
* split MIPS arch
* split ppc arch

see https://www.debian.org/releases/stretch/mips/ch02s01.html.en

related to https://github.com/elastic/beats/issues/19677 https://github.com/elastic/beats/issues/11750


go1.14/armel/rootfs/compilers.yaml
```diff
@@ -0,0 +1,17 @@
...
+    CXX: arm-linux-gnueabihf-g++
+  armv6:
+    # Using debian armel for ARMv6.
+    # This is the same as ARMv5 so it will use softfp instead of hardfp.
+    CC:  arm-linux-gnueabi-gcc
+    CXX: arm-linux-gnueabi-g++
+  armv5:
+    CC:  arm-linux-gnueabi-gcc
+    CXX: arm-linux-gnueabi-g++
+  arm64:
```
andrewkroh, line 15: I think arm64 and armv7 can be removed since these compilers are part of different images.

go1.14/armhf/rootfs/compilers.yaml
```diff
@@ -0,0 +1,17 @@
+---
+
+linux:
+  armv7:
+    CC:  arm-linux-gnueabihf-gcc
+    CXX: arm-linux-gnueabihf-g++
+  armv6:
```
andrewkroh, line 7: Only the armv7 compiler should be in this file.

go1.14/Makefile
```diff
@@ -1,4 +1,4 @@
-IMAGES := base main darwin arm mips ppc s390x
+IMAGES := base main amd64 darwin arm armel armhf mips ppc s390x
```
andrewkroh, line 1: Perhaps `arm` should be renamed to `arm64`.

go1.14/amd64/rootfs/compilers.yaml
```diff
@@ -0,0 +1,25 @@
+---
+
+windows:
+  386:
```
andrewkroh, line 4: The 386 compilers are no longer in this image IIUC so the `386` entries can be dropped. I think of this file as kind of a manifest so it's helpful to have it match up with what GOOS/GOARCH combos we expect the image to be used for.

go1.14/amd64/rootfs/compilers.yaml
```diff
@@ -0,0 +1,25 @@
...
+    CC:  i686-w64-mingw32-gcc
+    CXX: i686-w64-mingw32-g++
+  amd64:
+    CC:  x86_64-w64-mingw32-gcc
+    CXX: x86_64-w64-mingw32-g++
+
+darwin:
```
andrewkroh, line 11: darwin probably shouldn't have be in here since those are in the dedicated `darwin` image (perhaps I had this wrong in the `main` compilers.yaml).

go1.14/Makefile
andrewkroh: And `main` has effectively become the `i386` image.

andrewkroh: We'll have to update anything that uses the images to pull the right arch flavor so I guess if we're changing one (main -> amd64) then it won't be too much to rename arm and main...?

go1.14/amd64/rootfs/compilers.yaml
kuisathaverat: yep, I saw it, I reviewing all these `compilers.yaml` files

go1.14/Makefile
kuisathaverat: yes, we have to change the arch we pass when we launch the images, I start to find where this is in beats but I did not found it yet, Do you known where I have to touch to change it.

andrewkroh: It's at https://github.com/elastic/beats/blob/ded2e9912bb669e896b9c8d013dfabdb759023bd/dev-tools/mage/crossbuild.go#L180-L206. Journalbeat's magefile has a custom `ImageSelectorFunc` in order to use a different image due to debian7 + systemd issues. It will need updated too.

andrewkroh: We should probably make the image location configurable in Beats to allow for easier testing of the staged images (e.g. `export BEATS_GOLANG_CROSSBUILD_IMAGE=docker.elastic.co/observability-ci/golang-crossbuild`)

`BeatsCrossBuildImage = EnvOr("BEATS_GOLANG_CROSSBUILD_IMAGE", "docker.elastic.co/beats-dev/golang-crossbuild")`

https://github.com/elastic/beats/blob/7580c72445ecc6e97cab6fb7bcffc8063abf105b/dev-tools/mage/settings.go#L46

kuisathaverat:
I have checked supported architectures on CentOS, RedHat, Ubuntu, and Debian, the mips64 and ppc64 architectures are not supported by any of those distributions. On Debian, we have compilers for mips64 and ppc64 but do not have binary packages and it is not easy to prepare a bootstrap filesystem to compile all the packages we need. I think we can rid of mips64 and ppc64, these will be the supported architectures are:

|        Architecture        | Debian Designation |        Subarchitecture       |   Flavor   |   |
|:--------------------------:|:------------------:|:----------------------------:|:----------:|---|
| Intel x86-based            | i386               | default x86 machines         | default    |   |
|                            |                    | Xen PV domains only          | xen        |   |
| AMD64 & Intel 64           | amd64              |                              |            |   |
| ARM                        | armel              | Marvell Kirkwood and Orion   | marvell    |   |
| ARM with hardware FPU      | armhf              | multiplatform                | armmp      |   |
| 64bit ARM                  | arm64              |                              |            |   |
| 32bit MIPS (big-endian)    | mips               | MIPS Malta                   | 4kc-malta  |   |
|                            |                    | Cavium Octeon                | octeon     |   |
| 64bit MIPS (little-endian) | mips64el           | MIPS Malta                   | 5kc-malta  |   |
|                            |                    | Cavium Octeon                | octeon     |   |
|                            |                    | Loongson 3                   | loongson-3 |   |
| 32bit MIPS (little-endian) | mipsel             | MIPS Malta                   | 4kc-malta  |   |
|                            |                    | Cavium Octeon                | octeon     |   |
|                            |                    | Loongson 3                   | loongson-3 |   |
| Power Systems              | ppc64el            | IBM POWER8 or newer machines |            |   |
| 64bit IBM S/390            | s390x              | IPL from VM-reader and DASD  | generic    |   |

@andrewkroh @urso @elastic/observablt-robots

kuisathaverat:
I have made a test on how hard is to move down to Debian 8 (libc6-2.18), for amd64 and i386 is easy for the rest of the architectures we need to compile packages, I am not sure it worth it. Debian 7(libc6-2.13) need to compile a bunch of packages on any architecture. In my opinion, we can keep amd64 and i386 on Debian 8, and the rest should use Debian 9.

urso:
> I have made a test on how hard is to move down to Debian 8 (libc6-2.18), for amd64 and i386 is easy for the rest of the architectures we need to compile packages, I am not sure it worth it. Debian 7(libc6-2.13) need to compile a bunch of packages on any architecture. In my opinion, we can keep amd64 and i386 on Debian 8, and the rest should use Debian 9.

Can we reduce the amount of changes we introduce here? For the 7.9 release we only need go to be updated to go1.14.6

kuisathaverat:
@urso In that case, I will open another PR with only the change on the go version nothin else

kuisathaverat:
@andresrc @urso @andrewkroh we have to make a decision here, Do we want to split the Docker images in single architectures or not?

urso:
> we have to make a decision here, Do we want to split the Docker images in single architectures or not?

In general I'm +1 on splitting. What would be the disadvantages of doing this?

kuisathaverat:
> > we have to make a decision here, Do we want to split the Docker images in single architectures or not?
> 
> In general I'm +1 on splitting. What would be the disadvantages of doing this?

I do not see any, the images are easy to maintain, are smaller, and you use only the one you need.

andrewkroh:
+1 on splitting the images. Hopefully some of the layers can be shared between all of the images to minimize the amount of downloads/storage (but that's just an optimization and certainly not a requirement).

fearful-symmetry:
What's the status on this? It's blocking #71, which we'd like to resolve.

kuisathaverat:
> What's the status on this? It's blocking #71, which we'd like to resolve.

This was open in Jul, we have to re-work it over the new Go version, I will finish with https://github.com/elastic/golang-crossbuild/pull/72 and the nest thing to do is the split, give me a few days

urso:
>  we have to re-work it over the new Go version

Beats will not update to go 1.15 for the 7.11 release btw.

kuisathaverat:
@urso @fearful-symmetry Do we need it for 7.11? if so I will update this PR.

urso:
AFAICT there is no immedate need for this.

fearful-symmetry:
Yah, I think we're looking at 7.12 for the additional OS support.

kuisathaverat:
superseded by https://github.com/elastic/golang-crossbuild/pull/89

PR #75, opened 2021-01-28
Support docker buildx multiplatform

v1v:
### What

Use `docker-buildx` agents to build a multiplaform docker images for go1.14 and go1.15

Staging won't be needed anymore since `docker buildx --push` supports that feature now therefore no push explicitly but delegated to the makefile instead.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,6 +84,7 @@ pipeline {
 
 def buildAndPushImages(){
   dockerLogin(secret: "${env.DOCKER_REGISTRY_SECRET}", registry: "${env.REGISTRY}")
+  sh 'docker buildx create --use'
   withGoEnv(){
     dir("${env.BASE_DIR}"){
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -92,17 +82,12 @@ pipeline {
   }
 }
 
-def buildImages(){
+def buildAndPushImages(){
+  dockerLogin(secret: "${env.DOCKER_REGISTRY_SECRET}", registry: "${env.REGISTRY}")
+  sh 'docker buildx create --use'
```
v1v, line 87: Otherwise:

```
[2021-01-28T16:35:04.328Z] error: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

So I decided to delegate that particular configuration to the CI itself.

andresrc:
@v1v @kvch how this relates to #76 ?

kvch:
Mine is an alternative implementation. We have not decided which one to pick yet.

PR #80, opened 2021-03-09
Update Golang to 1.15.9

kvch:
It will break because the version has not been released yet.


michalpristas (approved):
i believe hash are correct

PR #87, opened 2021-05-13
Update darwin sdk and version

michel-laterman:
Update MacOS version + sdk to 10.13 and the corresponding Darwin version to 17.

- relates https://github.com/elastic/beats/issues/25346
- closes https://github.com/elastic/beats/issues/24193

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.REGISTRY}/observability-ci"
-    GO_VERSION = '1.15.12'
+    GO_VERSION = '1.16.3'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
@@ -40,12 +40,12 @@ pipeline {
         agent { label "${PLATFORM}"  }
         axes {
           axis {
-            name "MAKEFILE"
+            name 'MAKEFILE'
             values 'Makefile', 'Makefile.debian7', 'Makefile.debian8', 'Makefile.debian9', 'Makefile.debian10'
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.14', 'go1.15'
+            values 'go1.14', 'go1.15', 'go1.16'
           }
           axis {
             name 'PLATFORM'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.REGISTRY}/observability-ci"
-    GO_VERSION = '1.15.9'
+    GO_VERSION = '1.15.10'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.14', 'go1.15', 'go1.16'
+            values 'go1.16'
           }
           axis {
             name 'PLATFORM'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.16', 'go1.15'
+            values 'go1.16'
           }
           axis {
             name 'PLATFORM'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.REGISTRY}/observability-ci"
-    GO_VERSION = '1.16.3'
+    GO_VERSION = '1.16.4'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.REGISTRY}/observability-ci"
-    GO_VERSION = '1.15.10'
+    GO_VERSION = '1.15.12'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.16'
+            values 'go1.16' 'go1.15'
           }
           axis {
             name 'PLATFORM'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.16' 'go1.15'
+            values 'go1.16', 'go1.15'
           }
           axis {
             name 'PLATFORM'
--- a/README.md
+++ b/README.md
@@ -14,21 +14,21 @@ The base image used is Debian 9 (stretch) unless otherwise specified.
 
 ## Build Tags
 
-- `1.10.8-main`, `1.11.13-main`, `1.12.12-main`, `1.13.12-main`, `1.14.15-main`, `1.15.9-main` - linux/{amd64,386} and windows/{amd64,386}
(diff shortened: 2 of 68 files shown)


michel-laterman:
jenkins run tests

michel-laterman:
Currently the darwin build will fail; if the build debug flag is enabled and output is not piped to null we can see:
```
#6 113.4 [ 84%] Linking CXX static library ../libLLVMMC.a
#6 113.4 [ 84%] Built target LLVMMC
#6 113.4 CMakeFiles/Makefile2:28639: recipe for target 'projects/libtapi/tools/libtapi/CMakeFiles/libtapi.dir/rule' failed
#6 113.4 make[1]: *** [projects/libtapi/tools/libtapi/CMakeFiles/libtapi.dir/rule] Error 2
#6 113.4 Makefile:8361: recipe for target 'libtapi' failed
#6 113.4 make: *** [libtapi] Error 2
#6 113.4 + _exit
#6 113.4 + EC=2
#6 113.4 + '[' 2 -ne 0 ']'
#6 113.4 + test -z build.sh
#6 113.4 + echo ''
#6 113.4
#6 113.4 + echo 'exiting with abnormal exit code (2)'
#6 113.4 exiting with abnormal exit code (2)
#6 113.5 + test -n 1
#6 113.5 + '[' -n apple-libtapi ']'
#6 113.5 + rm -f 'build/*_built_successfully'
#6 113.5 + echo ''
#6 113.5
#6 113.5 + test build.sh = build.sh
#6 113.5 + check_for_bug_1242300
#6 113.5 + '[' -e /etc/issue ']'
#6 113.5 ++ grep -i ubuntu.13.10 /etc/issue
#6 113.5 + '[' '' ']'
```

ph:
@michel-laterman This is from the container?

michel-laterman:
Yes, there are currently 2 issues

1) debian9 (the base image) installs a version of clang that is too old to compile osxcross
2) the sdk for 10.13 is missing the libc++ headers

ph:
@urso What do you think about updating to a more recent version of debian? IIRC Debian 9 doesn't get any security updates anymore.

@michel-laterman with Debian10 do the two problems get resolved?

ph:
@andrewkroh would be good to have you comment on this.

michel-laterman:
@ph after recompressing (from bz2 to xz) a 10.14 SDK that another user has made (https://github.com/joseluisq/macosx-sdks/releases/tag/10.14) I was able to get osxcross compiling on debian10.

It seems that the releases that https://github.com/phracker/MacOSX-SDKs provides are broken for earler SDKs

urso:
GO 1.14 is EOL, and 1.15 will soon be EOL as well. This is why we switch to 1.16. I think we can remove 1.14 already.

> What do you think about updating to a more recent version of debian? IIRC Debian 9 doesn't get any security updates anymore.

For cross compiling to MacOS/Windows, the actual debian version should not matter.

For compiling on linux it matters unfortunately, and we need to keep debian7 in order to ensure our binaries continue to function with CentOS/RH7.
By accident I shared a Filebeat not build using the debian7 container 2 weeks ago, and the user was not able to run the binary on on RH7. RH7 goes EOL end of 2021, but we might have users that continue to use it even after.
The Beats-tester using a centos7 image should be able to discover potential problems (system linker complains on startup) if we update our debian build image. But debian10 might be too new I'm afraid.

> after recompressing (from bz2 to xz) a 10.14 SDK that another user has made

We are moving here from MacOS 10.12 to 10.14. As MacOS 10.13 is EOL since December 2020 I think this is fine. We still want to report this as a breaking change in the next Beats release.

Cross compilation can be kinda painful. Plus we have to jump through some hoops in order to sign our binaries. For the future we might consider to switch to native build environments. @v1v WDYT?

kuisathaverat:
>GO 1.14 is EOL, and 1.15 will soon be EOL as well. This is why we switch to 1.16. I think we can remove 1.14 already.

we have removed 1.14 a couple of days ago, we only build 1.15 and 1.16

>The Beats-tester using a centos7 image should be able to discover potential problems (system linker complains on startup) 

it also runs tests on CentOS 6

>Cross compilation can be kinda painful. Plus we have to jump through some hoops in order to sign our binaries. For the future we might consider to switch to native build environments.

We can switch now to build native environments for x86, ARM, and MacOS. We have workers for those architectures. However, this requires changes on the build system all the package parts must be replaced.

kuisathaverat:
> @ph after recompressing (from bz2 to xz) a 10.14 SDK that another user has made (https://github.com/joseluisq/macosx-sdks/releases/tag/10.14) I was able to get osxcross compiling on debian10.
> 
> It seems that the releases that https://github.com/phracker/MacOSX-SDKs provides are broken for earler SDKs

Update your PR with the changes on the master I have fixed all the stuff broken last week in the systemd dependencies. I seems like the new SDK does not work on old Libcs, it is fine we can use that new SDK only on Debian 10 and the old in the other versions of the cross-compiler Docker images. Then on the package build, we should choose the right Docker image depending on our target, on MacOS we do not have the issue related to libc compatibility is other OS.

go1.16/darwin/Dockerfile.tmpl
```diff
@@ -8,14 +8,20 @@ RUN \
...
+ARG OSXCROSS_SDK_URL=https://s3.amazonaws.com/beats-files/deps/MacOSX10.13.sdk.tar.xz
 ARG OSXCROSS_PATH=/usr/osxcross
-ARG OSXCROSS_REV=3034f7149716d815bc473d0a7b35d17e4cf175aa
-ARG SDK_VERSION=10.11
-ARG DARWIN_VERSION=15
-ARG OSX_VERSION_MIN=10.10
+ARG OSXCROSS_REV=8a716a43a72dab1db9630d7824ee0af3730cb8f9
+ARG SDK_VERSION=10.13
+ARG DARWIN_VERSION=17
+ARG OSX_VERSION_MIN=10.12
```
kuisathaverat, line 24:
```
{{if lt .DEBIAN_VERSION "10"}}
ARG OSXCROSS_SDK_URL=https://s3.amazonaws.com/beats-files/deps/MacOSX10.11.sdk.tar.xz
ARG OSXCROSS_PATH=/usr/osxcross
ARG OSXCROSS_REV=3034f7149716d815bc473d0a7b35d17e4cf175aa
ARG SDK_VERSION=10.11
ARG DARWIN_VERSION=15
ARG OSX_VERSION_MIN=10.10
{{ else }}
ARG OSXCROSS_SDK_URL=https://s3.amazonaws.com/beats-files/deps/MacOSX10.13.sdk.tar.xz
ARG OSXCROSS_PATH=/usr/osxcross
ARG OSXCROSS_REV=8a716a43a72dab1db9630d7824ee0af3730cb8f9
ARG SDK_VERSION=10.13
ARG DARWIN_VERSION=17
ARG OSX_VERSION_MIN=10.12
{{ end }}
```

kuisathaverat: this resolve the issue we will have the SDK 10.11 on Debian 7/8/9 and the SDK 10.13 on Debian 10

kuisathaverat:
now the SDK is downloaded and the osxcross compiled, but the `helloWorld test falied`

```
[2021-05-19T09:05:04.551Z] Step 15/21 : RUN cd /   && o64-clang helloWorld.c -o helloWorld   && file helloWorld   && file helloWorld | grep -c 'Mach-O 64-bit x86_64'

[2021-05-19T09:05:04.551Z]  ---> Running in 5c632da30120

[2021-05-19T09:05:04.551Z] /usr/osxcross/bin/x86_64-apple-darwin18-ld: error while loading shared libraries: libtapi.so.8svn: cannot open shared object file: No such file or directory
```

go1.16/darwin/Dockerfile.tmpl
```diff
@@ -36,11 +55,14 @@ RUN cd / \
   && file helloWorld \
   && file helloWorld | grep -c 'Mach-O 64-bit x86_64'
 
+# MacOSX10.14 SDK does not have 32bits compiler
+{{if ne .DEBIAN_VERSION "10"}}
 RUN cd / \
-  && o32-clang helloWorld.c -o helloWorld \
+  && o32-clang++ helloWorld.c -o helloWorld \
```
kuisathaverat, line 61:
```suggestion
  && o32-clang helloWorld.c -o helloWorld \
```

kuisathaverat:
ready to go

Jenkinsfile
```diff
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.16'
+            values 'go1.16', 'go1.15'
```
kuisathaverat, line 48:
```suggestion
            values 'go1.16'
```

README.md
```diff
@@ -73,3 +73,40 @@ GOARM, PLATFORM_ID, CC, and CXX.
...
+
+Then run:
+```
+../../target/SDK/tools/bin/pbzx -n Content | cpio -i
+cd ../..
+XCODEDIR=osxcross/build/tmp_<X> ./tools/gen_sdk_package.sh
+```
+
+The SDK should be in the working directory.
+The tmp dir can be safely deleted after this.
```
v1v, line 112: Do you think the upload to the bucket should be also added in this section?

urso: Yes, please.

michel-laterman: added

kuisathaverat (approved):

Issue #88, opened 2021-05-14
Unable to correct problems, you have held broken packages.


kuisathaverat:
there is an issue related with Debian 9 and some kind of issue with some packages dependencies

```
[2021-05-13T17:03:16.356Z] Step 4/12 : FROM ${REPOSITORY}/golang-crossbuild:${VERSION}-base${TAG_EXTENSION}
[2021-05-13T17:03:16.356Z]  ---> aa8b08ab436b
[2021-05-13T17:03:16.356Z] Step 5/12 : RUN     dpkg --add-architecture arm64     && apt-get update -qq     && apt-get install -qq -y       --no-install-recommends       --allow-unauthenticated         crossbuild-essential-arm64         crossbuild-essential-armel         crossbuild-essential-armhf         linux-libc-dev-arm64-cross         linux-libc-dev-armel-cross         linux-libc-dev-armhf-cross         librpm-dev:arm64         libc-dev:arm64         libpopt-dev:arm64         linux-libc-dev:arm64         libxml2-dev:arm64         libxml2:arm64         libicu-dev:arm64         libicu57:arm64         icu-devtools:arm64         libsystemd-dev:arm64         librpm3:arm64          librpmio3:arm64         librpmbuild3:arm64         librpmsign3:arm64         libsqlite3-dev:arm64         libnss3:arm64         libsqlite3-0:arm64     && rm -rf /var/lib/apt/lists/*
[2021-05-13T17:03:16.356Z]  ---> Running in 51a7b8393af1
[2021-05-13T17:03:22.941Z] E: Unable to correct problems, you have held broken packages.
[...]

PR #89, opened 2021-05-14
fix: systemd dependencies for ARM and MIPS

kuisathaverat:
* fix: add emdebian.org repo
* fix: systemd dependencies for ARM

closes https://github.com/elastic/golang-crossbuild/issues/88

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
           }
           axis {
             name 'GO_FOLDER'
-            values 'go1.14', 'go1.15', 'go1.16'
+            values 'go1.16'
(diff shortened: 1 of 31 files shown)


v1v (approved):

kuisathaverat:
🤔 Should I fix it for older Docker images 1.14 and 1.15, I star thinking we should rid of old Go versions and keep only the supported ones.

kuisathaverat:
I have removed Go 1.14 and Go 1.15, we do not use them anymore, and we have generated Docker images in the repository for those cases in case we need them.

cachedout (approved):

PR #99, opened 2021-05-26
chore: remove unsupported builds

kuisathaverat:
This repository has older versions of Go out of support, in each one of those we have configuration files to build docker images based on several distributions some of them also out of support. The build of those images is broken and we will not invest time to fix those build because does not worth the time. So due to we will not fix the issues because we do not use those images anymore, we should rid of them to avoid people think that those builds should work.
We will still have the Docker images in the Docker registry so we can still use old Docker images generated and we keep a tag to the current status of the repository https://github.com/elastic/golang-crossbuild/releases/tag/1.15


cachedout (approved):

Makefile
```diff
@@ -1,4 +1,4 @@
-TARGETS=go1.10 go1.11 go1.12 go1.13 go1.14 go1.15 go1.16
+TARGETS=go1.15 go1.16
```
urso, line 1: @kvch shall we keep go1.13 and go1.14? So we can test the functionbeat build with the version used by GCP?

v1v: IIIUC, those docker images are available anyway since they were built already but since they are not patched anymore we don't need to keep their configuration

v1v (approved):

Makefile
urso: Sounds good.

urso (approved):

mdelapenya (approved):
Dracarys!

dracarys

PR #142, opened 2021-12-13
Create release notes

v1v:
### What

Automate the release with `gren` to generate the changelog and `gh` to publish the release notes.

### Why

Unfortunately `gren changelog` and `gren release` don't have the some outcome, therefore all the templating we added in the `.grenrc` file is not used when running `gren release`.

### Test

image

image

Bear in mind there are two entries for the 1.17.3 release, since there was already some details in place, hopefully this should not happen with the future releases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ _obj
 Dockerfile
 .status.*
 *.bck
+CHANGELOG.md
--- a/.ci/scripts/release-notes.sh
+++ b/.ci/scripts/release-notes.sh
@@ -4,4 +4,4 @@ set +x
 
 GREN_GITHUB_TOKEN=${GREN_GITHUB_TOKEN:?"missing GREN_GITHUB_TOKEN"}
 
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ _obj
 Dockerfile
 .status.*
 *.bck
+CHANGELOG.md
```
cachedout, line 10: Is the idea here that the changelog will _only_ be present in release artifacts? It would be nice if we could find a way to keep a copy in the repo itself as well IMHO.

v1v: This file was intended to use to bypass the issue with `gren release`. So far the release notes are stored in the GitHub releases. 
We can enable the `CHANGELOG.md` within this repo in a follow up, what do you think?

cachedout: Fine to do in a follow-up. Thanks.

cachedout (approved):

Issue #173, opened 2022-04-13
debian 9: unmet dependencies


v1v:
```
The following packages have unmet dependencies:
[2022-04-13T15:50:48.789Z]  libelf1:mips64el : Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  libnss3:mips64el : Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  librpm-dev:mips64el : Depends: libz-dev:mips64el
[2022-04-13T15:50:48.789Z]  librpmbuild3:mips64el : Depends: libmagic1:mips64el (>= 5.12) but it is not going to be installed

[2022-04-13T15:50:48.789Z]  librpmio3:mips64el : Depends: liblzma5:mips64el (>= 5.1.1alpha+20120614) but it is not going to be installed

[2022-04-13T15:50:48.789Z]                       Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed

[2022-04-13T15:50:48.789Z]  libxml2:mips64el : Depends: liblzma5:mips64el (>= 5.1.1alpha+20120614) but it is not going to be installed

[2022-04-13T15:50:48.789Z]                     Depends: zlib1g:mips64el (>= 1:1.2.3.3) but it is not going to be installed

[2022-04-13T15:50:48.789Z] E: Unable to correct problems, you have held broken packages.

[2022-04-13T15:50:48.990Z] The command '/bin/sh -c apt install -y         libelf1:mips64el         libicu-dev:mips64el         libicu57:mips64el         librpm-dev:mips64el         librpm3:mips64el         librpmio3:mips64el         librpmbuild3:mips64el         librpmsign3:mips64el         libxml2-dev:mips64el         libsqlite3-dev:mips64el         libnss3:mips64el         libsqlite3-0:mips64el         libxml2:mips64el         libsqlite3-0:mips64el' returned a non-zero code: 100

[2022-04-13T15:50:48.990Z] make[1]: *** [../Makefile.common:15: build] Error 100

[2022-04-13T15:50:48.990Z] make[1]: Leaving directory '/var/lib/jenkins/workspace/ats_golang-crossbuild-mbp_PR-171/src/github.com/elastic/golang-crossbuild/go/mips'

[2022-04-13T15:50:48.990Z] make: *** [Makefile.debian9:9: build] Error 1

[...]

PR #174, opened 2022-04-15
Correct unmet dependencies on mips, mips32, ppc and s390x #173

amannocci:
Resolve issue #173 by downgrading [zlib1g](https://packages.debian.org/stretch/zlib1g) and [liblzma5](https://packages.debian.org/stretch/liblzma5) to a previous version.

Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

### DIFF
--- a/go/mips/Dockerfile.tmpl
+++ b/go/mips/Dockerfile.tmpl
@@ -22,7 +22,7 @@ RUN apt install -qq -y \
 
 {{ if eq .DEBIAN_VERSION "9" }}
 # librpm-dev
-RUN apt install -y \
+RUN apt install -y --allow-downgrades \
         libelf1:mips64el \
         libicu-dev:mips64el \
         libicu57:mips64el \
@@ -36,7 +36,9 @@ RUN apt install -y \
         libnss3:mips64el \
         libsqlite3-0:mips64el \
         libxml2:mips64el \
-        libsqlite3-0:mips64el
+        libsqlite3-0:mips64el \
+        liblzma5:amd64=5.2.2-1.2+b1 \
+        zlib1g:amd64=1:1.2.8.dfsg-5
 
 # libsystemd-dev
 # RUN apt install -y \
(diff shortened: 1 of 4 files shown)


v1v (commented):
Thanks Adrien for this PR 👍 

Couple of questions:

1) The platform in the package were pointing to amd64 rather than the specific platform, see the suggestions.
2) `allow-downgrades` is not recommended as far as I see in the docs

>--allow-downgrades
This is a dangerous option that will cause apt to continue without prompting if it is
doing downgrades. It should not be used except in very special situations. Using it
can potentially destroy your system!

Is it possible to add the specific dependencies and versions without the flag?

kuisathaverat:
I'd avoid using fixed versions because we will lose security updates. When you  install crosscompile packages the dependencies are not resolved so you have to add all missing dependencies to the list of packages to install this is way it fails in first instance.

```
[2022-04-13T15:50:48.789Z]  libelf1:mips64el : Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  libnss3:mips64el : Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  librpm-dev:mips64el : Depends: libz-dev:mips64el
[2022-04-13T15:50:48.789Z]  librpmbuild3:mips64el : Depends: libmagic1:mips64el (>= 5.12) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  librpmio3:mips64el : Depends: liblzma5:mips64el (>= 5.1.1alpha+20120614) but it is not going to be installed
[2022-04-13T15:50:48.789Z]                       Depends: zlib1g:mips64el (>= 1:1.1.4) but it is not going to be installed
[2022-04-13T15:50:48.789Z]  libxml2:mips64el : Depends: liblzma5:mips64el (>= 5.1.1alpha+20120614) but it is not going to be installed
[2022-04-13T15:50:48.789Z]                     Depends: zlib1g:mips64el (>= 1:1.2.3.3) but it is not going to be installed
```

we need to add the new dependencies `libmagic1:mips64el`, `liblzma5:mips64el`, and `zlib1g:mips64el`, changing the architecture as it is needed for each architecture.

kuisathaverat:
superseded by https://github.com/elastic/golang-crossbuild/pull/175

PR #182, opened 2022-04-20
feat: job to compile the LLVM Apple version

kuisathaverat:
It adds a new job to compile the LLVM Apple version and push the Docker image

PR #184, opened 2022-04-20
chore: move golang crossbuild job to a dedicated folder

kuisathaverat:
move golang crossbuild job to a dedicated folder


v1v (approved):
But the folder definition is somewhere else, so this should not be merged until the other PR, right?

kuisathaverat:
>But the folder definition is somewhere else, so this should not be merged until the other PR, right?

it is merged a long with https://github.com/elastic/golang-crossbuild/pull/182

PR #192, opened 2022-04-27
chore: remove Go version from the LLVM image tag

kuisathaverat:
The Go version is not relevant for the LLVM Docker image, it only depends on the Debian version we are using.

### DIFF
--- a/.ci/llvm-apple.groovy
+++ b/.ci/llvm-apple.groovy
@@ -44,7 +44,7 @@ pipeline {
     stage('Check changes'){
       when {
         anyOf {
-          changelog '/go/llvm-apple/**'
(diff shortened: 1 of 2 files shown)


v1v (approved):

v1v (approved):
Will this work for the very first builds?

kuisathaverat:
>Will this work for the very first builds?

it is complicated, it depends on the `base` Docker image, so it will not work if there is no a PR with the changes for the new Debian version. Resolve this is not the scope of this PR.

To resolve the dependency of the base Docker image the LLVM Apple Docker image should move all the login in the base Docker image to the LLVM Apple Docker image

PR #216, opened 2022-07-06
fix: use bash to run the commands

kuisathaverat:
use bash to run the commands

### DIFF
--- a/go/base/rootfs/entrypoint.go
+++ b/go/base/rootfs/entrypoint.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package main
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

go/base/rootfs/entrypoint.go
```diff
@@ -154,7 +155,7 @@ func loadCompilerSettings(goos, arch string, env map[string]string) error {
 }
 
 func execBuildCommand(env map[string]string) error {
-	cmd := exec.Command("sh", "-c", buildCommand)
+	cmd := exec.Command("bash", "-c", "-l", buildCommand)
```
v1v, line 158: To help with:

> -l        Make bash act as if it had been invoked as a login shell (see INVOCATION below).

I didn't know that flag :)

kuisathaverat:
it does not resolve the issue, indeed remove some env vars more

PR #220, opened 2022-07-29
Disable automatic APT updates and clear cache

florianl:
- Automatic APT updates can introduce an overhead when running the docker image. This results in additional noise. So disabling it shifts the focus on the task of the docker container. In addition it helps to create reproducible builds as the dependencies do not change automatically without intention. 

- Clearing the APT cache reduces the size of the docker image.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
     STAGING_IMAGE = "${env.DOCKER_REGISTRY}/observability-ci"
-    GO_VERSION = '1.18.4'
+    GO_VERSION = '1.18.5'
     BUILDX = "1"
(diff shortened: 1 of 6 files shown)


AndersonQ (approved):

ph (approved):

ph:
@mergifyio update

ph:
interesting error on the CI, we don't build explicitely for `armel` platform I am confused why the cross build docker would require them? @andrewkroh Do you remember what it's needed here?

andrewkroh:
> interesting error on the CI, we don't build explicitely for armel platform I am confused why the cross build docker would require them?

I'm parsing this as "we don't build [beats/agent] explicitly for armel platform . . ." If someone wanted to target linux/arm (v5 or v6) then the image would need to exist. I'd be curious to see how often the armel image is downloaded if we have those metrics from our docker registry. My guess is that most arm usage is on arm (v7) (aka armhf) and arm64.

florianl:
Closing this PR, as I'm not able to fix the CI issue atm.

Issue #245, opened 2022-11-07
libc6-dev : Depends: libc6 (= 2.31-13+deb11u4)


v1v:
Debian 11 docker images for golang-crossbuild are failing with the below error
```
[2022-11-07T04:16:39.605Z] #11 6.158  libc6-dev : Depends: libc6 (= 2.31-13+deb11u4) but 2.31-13+deb11u5 is to be installed
```

v1v:
<img width="849" alt="image" src="https://user-images.githubusercontent.com/2871786/200326700-0adc3b1f-9825-42f3-89e2-24334cdf6a92.png">

vs 

<img width="1411" alt="image" src="https://user-images.githubusercontent.com/2871786/200326759-b9839a16-54d0-4c4b-9025-8261905efd43.png">

PR #246, opened 2022-11-07
fix libc6-dev : Depends: libc6 (= 2.31-13+deb11u4) but 2.31-13+deb11u5

v1v:
### What

Debian 11 fails when installing the dependency libc6 since it requires a downgraded version instead. 

### Errors

image

Closes https://github.com/elastic/golang-crossbuild/issues/245

### DIFF
--- a/go/base/Dockerfile.tmpl
+++ b/go/base/Dockerfile.tmpl
@@ -25,6 +25,9 @@ RUN apt-get -o Acquire::Check-Valid-Until=false update -y --no-install-recommend
             python3-pip \
             python3 \
 {{- end }}
+{{- if (eq .DEBIAN_VERSION "11")}}
+            libc6=2.31-13+deb11u4 --allow-downgrades \
+{{- end }}
         && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


go/base/Dockerfile.tmpl
```diff
@@ -24,6 +24,9 @@ RUN apt-get -o Acquire::Check-Valid-Until=false update -y --no-install-recommend
             python3-venv \
             python3-pip \
             python3 \
+{{- end }}
+{{- if (eq .DEBIAN_VERSION "11")}}
+            libc6=2.31-13+deb11u4 --allow-downgrades \
```
v1v, line 29: this dependency is needed when installing `build-essentials`: https://packages.debian.org/bullseye/devel/build-essential

kuisathaverat, line 29:
```suggestion
# FIXME remove when the libc6-dev is updated to the new libc6
            libc6=2.31-13+deb11u4 --allow-downgrades \
```

v1v:
@mergifyio backport 1.18

v1v:
@Mergifyio backport 1.18

PR #262, opened 2023-02-06
automate bump golang version

v1v:
## What does this PR do?

Automate the golang bump within the project itself using [updatecli](https://www.updatecli.io/) and github actions

## Why is it important?

Remove the dependency with Jenkins and the definition of what to do in [apm-pipeline-library](https://github.com/elastic/apm-pipeline-library/blob/a3850732518748f2e71815c093d58cb071d0d8a3/.ci/.bump-go-release-version.yml#L33-L40)

Use updatecli that provides the tooling for bumping versions

## Further details

There will be one GitHub workflow per active Golang minor version.

### DIFF
--- a/.github/workflows/bump-golang-1.18.yml
+++ b/.github/workflows/bump-golang-1.18.yml
@@ -2,6 +2,7 @@
 name: Bump golang version 1.18
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: '0 22 * * 6'
 
--- a/.github/workflows/bump-golang-1.18.yml
+++ b/.github/workflows/bump-golang-1.18.yml
@@ -3,7 +3,7 @@ name: Bump golang version 1.18
 
 on:
   schedule:
-    - cron: '0 21 * * 6'
+    - cron: '0 22 * * 6'
 
 permissions:
   pull-requests: write
--- a/.github/workflows/bump-golang-1.18.yml
+++ b/.github/workflows/bump-golang-1.18.yml
@@ -1,24 +1,21 @@
 ---
-name: Bump golang version
+name: Bump golang version 1.18
 
 on:
   schedule:
-    - cron: '0 20 * * 6'
+    - cron: '0 21 * * 6'
 
 permissions:
   pull-requests: write
   contents: write
 
-env:
-  BRANCH_TO_BE_UPDATED: main
-
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/bump-golang-1.18.yml
```diff
@@ -0,0 +1,21 @@
+---
```
kuisathaverat, line 1: Why split it into three workflows instead of one matrix workflow?

v1v: i initially did it, but I somehow found the `filename` with the go-version much appealing and easy to track that having one workflow with the matrix list. It's just a cosmetic change, but files with simpler workflows might help with the maintenance

kuisathaverat: Edit one alternative for the matrix vs create/delete new files. 
If the matrix file contains the version and the SHA we can get rid of part of the update script and load the SHA and version from that file too.

```yaml
versions:
  - name: 1.20.0
    sha: 98579582350y183425198375098370597340597340
  - name: 1.19.0
    sha: 98579582350y183425198375098370597340597340
  - name: 1.18.0
    sha: 98579582350y183425198375098370597340597340
```

Anyways it looks like more implementation effort right now so let's live with three workflows

kuisathaverat (approved):

.github/workflows/bump-golang.yml
```diff
@@ -0,0 +1,24 @@
+---
+name: Bump golang version
+
+on:
+  schedule:
```
v1v, line 5:
```suggestion
  workflow_dispatch:
  schedule:
```

.github/workflows/bump-golang-1.19.yml
```diff
@@ -0,0 +1,21 @@
+---
+name: Bump golang version 1.19
+
+on:
+  schedule:
```
v1v, line 5:
```suggestion
  workflow_dispatch:
  schedule:
```

.github/workflows/bump-golang-1.18.yml
v1v: That's a good approach, though it might require to change a bit more. For now as you said, let's keep it simple to support the current implementation. This might be changed by the ingest team in the future, maybe.

PR #264, opened 2023-02-07
action: fix relative path

v1v:
### What

* use opentelemetry exporter
* change names
* fix relative path with composite actions

### DIFF
--- a/.github/workflows/bump-golang-1.18.yml
+++ b/.github/workflows/bump-golang-1.18.yml
@@ -1,5 +1,5 @@
 ---
-name: Bump golang version 1.18
+name: bump-golang-1.18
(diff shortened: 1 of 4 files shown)


.github/workflows/bump-golang-1.18.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: .github/workflows/bump-golang
+      - uses: ./.github/workflows/bump-golang
```
v1v, line 18: This is the fix!

amannocci (approved):

PR #266, opened 2023-02-13
updatecli: use condition to fix shell action idempotence

v1v:
Shell actions in `updatecli` require some specific implementation see https://www.updatecli.io/docs/plugins/resource/shell/#_shell_target

Hence, the condition will ensure no changes need to be applied in the latest release matches the one in the file `go/Makefile.common`.

Fixes https://github.com/elastic/golang-crossbuild/issues/265


reakaleek (approved):

v1v:
@mergifyio backport 1.19

Issue #231, opened 2022-09-08
windows32 builds did not work until I set up `CGO_CFLAGS="-D_WIN32_WINNT=0x0600"`


igor-hnizdo:
I had weird issues with builds for windows32; I kept getting

```
undefined reference to `SleepConditionVariableCS'
```

and such.

When I used

```
CGO_CFLAGS="-D_WIN32_WINNT=0x0600"
```

everything started to work. As that seems to be Windows version Vista, where that particular API has been introduced in kernel32.

I copied the line from here

https://github.com/karalabe/xgo/blob/c5ccff8648a77686d274c1984e1cd6319736f5cb/docker/base/build.sh#L442

It should be at least documented somewhere. Or maybe, even pre-filled (I don't think people will build binaries for pre-Vista Windows, but who knows)

kuisathaverat:
We have dropped support to win32, but yep we should put it in the README, and stop releasing i386 Docker images

v1v:
https://github.com/elastic/golang-crossbuild/pull/304 has been merged, so docs are now reflecting that windows/i386 is not something supported even though if docker images are available.

I'll close this issue now

PR #304, opened 2023-06-21
docs: remove windows 32 references

v1v:
Even though if they are built they are not supported. So for the time being this will help consumers so they know what to expect when using the unsupported win32 platform.

part of https://github.com/elastic/golang-crossbuild/issues/231

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The `armel`, `mips`, `ppc` and `s390x` platforms are only supported from `Debian
 
 `mips32` is not available in `Debian 11`.
 
-`win32` has been dropped from Golang `1.19.11`/`1.20.6` onwards.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,6 +12,8 @@ The `armel`, `mips`, `ppc` and `s390x` platforms are only supported from `Debian
 
 `mips32` is not available in `Debian 11`.
 
+`win32` has been dropped from Golang `1.19.11`/`1.20.6` onwards.
```
v1v, line 15:
```suggestion
`windows/386` has been dropped from Golang `1.19.11`/`1.20.6` onwards.
```

kuisathaverat (approved):

Issue #311, opened 2023-08-17
debian-11 failed when building the docker image `1.21.0-main-debian11`


v1v:
```
[2023-08-16T10:22:31.447Z] >> Building docker.elastic.co/beats-dev/golang-crossbuild:1.21.0-main-debian11
[2023-08-16T10:22:31.949Z] #0 building with "default" instance using docker driver
[2023-08-16T10:22:31.949Z] 
[2023-08-16T10:23:45.480Z] #9 [ 4/10] RUN apt install -y --no-install-recommends --allow-unauthenticated         libsystemd-dev
[2023-08-16T10:23:45.781Z] #9 0.299 
[2023-08-16T10:23:45.781Z] #9 0.299 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
[2023-08-16T10:23:45.781Z] #9 0.299 
[2023-08-16T10:23:46.784Z] #9 0.306 Reading package lists...
[2023-08-16T10:23:47.085Z] #9 1.442 Building dependency tree...
[2023-08-16T10:23:47.085Z] #9 1.714 Reading state information...
[2023-08-16T10:23:47.186Z] #9 1.843 Some packages could not be installed. This may mean that you have
[2023-08-16T10:23:47.186Z] #9 1.843 requested an impossible situation or if you are using the unstable
[2023-08-16T10:23:47.186Z] #9 1.843 distribution that some required packages have not yet been created
[2023-08-16T10:23:47.186Z] #9 1.843 or been moved out of Incoming.
[2023-08-16T10:23:47.186Z] #9 1.843 The following information may help to resolve the situation:
[2023-08-16T10:23:47.186Z] #9 1.843 
[2023-08-16T10:23:47.186Z] #9 1.843 The following packages have unmet dependencies:
[2023-08-16T10:23:47.186Z] #9 1.920  libsystemd-dev : Depends: libsystemd0 (= 247.3-7+deb11u2)
[2023-08-16T10:23:47.286Z] #9 1.946 E: Unable to correct problems, you have held broken packages.
[2023-08-16T10:23:47.286Z] #9 ERROR: process "/bin/sh -c apt install -y --no-install-recommends --allow-unauthenticated         libsystemd-dev" did not complete successfully: exit code: 100
[2023-08-16T10:23:47.286Z] ------
[2023-08-16T10:23:47.286Z]  > [ 4/10] RUN apt install -y --no-install-recommends --allow-unauthenticated         libsystemd-dev:
[...]

PR #312, opened 2023-08-21
debian11: solve libsystemd-dev dependencies

v1v:
See https://packages.debian.org/bullseye/libsystemd-dev

image

Closes https://github.com/elastic/golang-crossbuild/issues/311


v1v:
I fixed the original error, though there is another failure afterwards when building the `arm` docker image, see https://github.com/elastic/golang-crossbuild/issues/311#issuecomment-1685892808

v1v:
https://github.com/elastic/golang-crossbuild/pull/316 is the one fixing it

PR #148, opened 2022-01-11
Provide npcap distribution internally

v1v:
### What

Provide the npcap distribution in an internal docker image, for such, it uses the internal docker registry, the main docker image and copy the file to `/installer` folder.

### Actions

- [x] Verifies no security disclosure.
- [x] Add `npcap` version in the tag image.

#### Build

image

image

#### Push

image

#### Pull

image

#### Files

image

PR #315, opened 2023-09-10
bump npcap version to 1.76

efd6:
This bumps the version of the Npcap OEM installer to v1.76. The artifact has been placed in the private store.

Please take a look.

Query: What backports will be required for this change?

For elastic/beats#36539


efd6:
Test failure appears to be unrelated.

v1v:
> Query: What backports will be required for this change?

It depends on the consumers. Golang-crossbuild follows the Golang versioning, only one minor version is supported. For instance, `1.21.1` is the latest so `1.20` is the last supported minor.

v1v (approved):

efd6:
OK, thanks. This will only be for children of beats main (with no backports), which is at go1.20.8, so no backport other that go1.20 will be needed.

pazone (approved):

v1v:
> OK, thanks. This will only be for children of beats main (with no backports), which is at go1.20.8, so no backport other that go1.20 will be needed.

I'm afraid I didn't explain this correctly. `golang-crossbuild:1.20.8` has been already released. Hence it cannot be changed unless using the above-mentioned branch strategy.

If this PR is merged and backported to `1.20` then this change won't be available until `1.20.9`. Then, Beats will need to use that version or above.

If you need this change to be in `golang-crossbuild:1.20.8` then, it's required to branch off `1.20.8.x` and backport this change so the docker images for `golang-crossbuild:1.20.8` can be recreated.

**NOTE**: this is the reason we created https://github.com/elastic/golang-crossbuild/tree/1.17.5.x , so https://github.com/elastic/golang-crossbuild/pull/148 could be available in `1.17.5`

efd6:
OK. That's a lot of rigmarole. I'll see what I can do.

v1v:
> OK. That's a lot of rigmarole. I'll see what I can do.

Thanks for trying to understand my writing!  To help with, let me ask you this:
* do you need this change in `1.20.8`.
* O can you wait for `1.20.9` or `1.21.2`?

If the former, then we need all that craziness. If the latter, then there is a need to request the Beats team to use `1.20.9` or `1.21.2` when golang-crossbuild provides those versions (likely in 3 weeks or so)

efd6:
I don't know what the time frame for a go1.20.9 is. Beats doesn't build with go1.21 yet. I guess it is likely to be before 8.11, so we could wait. It's not a wildly friendly system though.

v1v:
> I guess it is likely to be before 8.11, so we could wait. It's not a wildly friendly system though.

I can happily help with and make some notes so whoever needs to do this in the future can follow them. For doing so, let me know and merge this PR please :)

efd6:
I'm unable to merge this; it is still waiting for a review from @alexsapran.

alexsapran (approved):

alexsapran:
> I'm unable to merge this; it is still waiting for a review from @alexsapran.

Sorry for that, I saw @pazone gave the 👍🏼 and I thought it was enough.

v1v:
@mergifyio backport 1.20.8.x

PR #318, opened 2023-09-20
bump npcap version to 1.76

efd6:
This bumps the version of the Npcap OEM installer to v1.76. The artifact has been placed in the private store.

(I don't think I've done this right, but it's entirely unclear what is actually expected)

Please take a look.

For elastic/beats#36539


v1v (approved):

efd6:
Yeah, this was to address https://github.com/elastic/golang-crossbuild/pull/315#issuecomment-1727117452

PR #321, opened 2023-09-25
docs: how to backport ncap to an existing golang-crossbuild version

v1v:
Trying to explain in the existing docs how the ncap could be backported to an existing golang-crossbuild version.

https://github.com/elastic/golang-crossbuild/pull/315#issuecomment-1723228820 is the one explaining these steps tool.


efd6 (approved):
Thanks

PR #326, opened 2023-10-26
Migrate golang-crossbuild pipeline

oakrizan:
In scope of [1695](https://github.com/elastic/ingest-dev/issues/1695) migrated golang-crossbuild job


jlind23 (dismissed):

.buildkite/scripts/publish.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -euo pipefail
+
+source .buildkite/scripts/common.sh
+
+MAKEFILE=${1}
+
+check_is_arm
+add_bin_path
+make -C go -f "${MAKEFILE}" push"${is_arm}"
```
mrodm, line 11: In Jenkins, this command is retried 3 times in case it fails.
Should we add this ? 

```suggestion
retry 3 make -C go -f "${MAKEFILE}" push"${is_arm}"
```

.buildkite/scripts/post-release.sh
```diff
@@ -0,0 +1,32 @@
...
+}
+
+tag_commit() {
+  echo "Tagging commit ${BUILDKITE_COMMIT}"
+  git tag -a -m "${BUILDKITE_COMMIT}" "${TAG}"
+}
+
+git_push_with_auth() {
+  echo "Pushing tag ${TAG}"
+  retry 3 git push https://${GITHUB_USERNAME_SECRET}:${GITHUB_TOKEN_SECRET}@github.com/elastic/golang-crossbuild.git ${TAG}
```
mrodm, line 21: This command does not look like exactly the same as in Jenkins:
https://github.com/elastic/golang-crossbuild/blob/b66e5a57400436f8fb206ad2742a394ffb9d0104/Jenkinsfile#L249C7-L249C19

In jenkins it is used `--force` parameter for pushing the tag.

https://github.com/elastic/apm-pipeline-library/blob/5800f4a0e2c6b29ff642b29279cb4b6eb9095cbc/vars/gitCreateTag.groovy#L31
https://github.com/elastic/apm-pipeline-library/blob/5800f4a0e2c6b29ff642b29279cb4b6eb9095cbc/vars/gitPush.groovy#L28

Should it be something like this ?
```suggestion
  retry 3 git push https://${GITHUB_USERNAME_SECRET}:${GITHUB_TOKEN_SECRET}@github.com/elastic/golang-crossbuild.git --tags --force
```

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,29 @@
...
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" && ("$BUILDKITE_STEP_KEY" =~ ^build || "$BUILDKITE_STEP_KEY" =~ ^publish)]]; then
+    export DOCKER_USERNAME_SECRET=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
+    export DOCKER_PASSWORD_SECRET=$(retry 5 vault kv get -field password "${DOCKER_REGISTRY_SECRET_PATH}")
+    docker login -u "${DOCKER_USERNAME_SECRET}" -p "${DOCKER_PASSWORD_SECRET}" "${DOCKER_REGISTRY}" 2>/dev/null
+fi
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" && "$BUILDKITE_STEP_KEY" == "post-release" ]]; then
+    export GITHUB_TOKEN_SECRET=$(retry 5 vault kv get -field token ${GITHUB_TOKEN_VAULT_PATH})
+    export TAG_EXISTS=$(tagExists ${GITHUB_TOKEN_VAULT_PATH} v${GOLANG_VERSION})
```
mrodm, line 26: Since this (TAG_EXISTS) is not a secret, could this calculation be moved to the `post-release.sh` script ?

mrodm, line 26: `tagExists` function expects as first parameter the secret not the vault path.

.buildkite/scripts/common.sh
```diff
@@ -0,0 +1,144 @@
...
+  echo "Deleting temporary files..."
+  rm -rf ${WORKSPACE}/${TMP_FOLDER}.*
+  echo "Done."
+}
+
+tagExists() {
+  local ghSecret=$1
+  local tag=$2
+  local url=https://api.github.com/repos/elastic/${REPO}/releases/tags/${tag}
+  local status=$(curl -s -o /dev/null -w "%{http_code}" -u ${ghSecret}:x-oauth-basic ${url})
```
mrodm, line 129: Instead of using `ghSecret` here and being passed the secret as a parameter, what about using directly GITHUB_TOKEN_SECRET in the command ? That secret should be available (and redacted) thanks to the pre-command hook

```suggestion
  local status=$(curl -s -o /dev/null -w "%{http_code}" -u ${GITHUB_TOKEN_SECRET}:x-oauth-basic ${url})
```

That would make that tagExists just needs one parameter, the tag.

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,118 @@
...
+  BUILDX: 1
+
 steps:
-  - label: "Example test"
-    command: echo "Hello!"
+  - group: "Staging"
+    key: "staging"
+
+    steps:
+      - label: ":unix: Staging / Ubuntu X86_64 - {{matrix.makefile}}"
```
mrodm, line 18: unix is not part of the emoji list available in Buildkite
https://github.com/buildkite/emojis#emoji-reference

What about using :linux:? 

```suggestion
      - label: ":linux: Staging / Ubuntu X86_64 - {{matrix.makefile}}"
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,118 @@
...
+            makefile:
+              - "Makefile"
+              - "Makefile.debian7"
+              - "Makefile.debian8"
+              - "Makefile.debian9"
+              - "Makefile.debian10"
+              - "Makefile.debian11"
+              - "Makefile.debian12"
+
+      - label: ":unix: Staging / Ubuntu ARM - {{matrix.makefile}}"
```
mrodm, line 41: unix is not part of the emoji list available in Buildkite
https://github.com/buildkite/emojis#emoji-reference

What about using :linux:? 
```suggestion
      - label: ":linux: Staging / Ubuntu ARM - {{matrix.makefile}}"
```

mrodm: There are other usages of this in 
https://github.com/elastic/golang-crossbuild/pull/326/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R68
https://github.com/elastic/golang-crossbuild/pull/326/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R88

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
oakrizan: Yes, it is.

.buildkite/scripts/buildx.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e
+set +x
+
+BUILDPLATFORM=${BUILDPLATFORM:-"linux/amd64,linux/arm64"}
+
+#docker run --privileged --rm tonistiigi/binfmt --install all
```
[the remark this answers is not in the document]
oakrizan: I have just copied the script, that is originally used in Jenkins, so that after .ci folder deletion I could simply change the related path. Not sure why original contains the commented code, but I've just kept it for a meantime.

.buildkite/scripts/post-release.sh
oakrizan: I was suspicious about force pushing tags, since that might lead to issues like conflicts between remote-local tags etc. And ideally tag always should point in a same commit and should not be moved to avoid confusion. I understand that there is additional check in post-release script, but anyways I'd try to avoid force pushing.

.buildkite/scripts/common.sh
dliappis, line 129: Especially when pulling data from github it's recommended to enable retries as it can frequently fail. We can either use curl's native [`--retry`](https://everything.curl.dev/usingcurl/downloads/retry) or the `retry()` bash helper included in this script.

.buildkite/scripts/post-release.sh
```diff
@@ -0,0 +1,32 @@
...
+}
+
+if [ ${TAG_EXISTS} ]; then
```
dliappis, line 24: Looking at the declaration of `TAG_EXISTS` [earlier in the code](https://github.com/elastic/golang-crossbuild/pull/326/files#diff-68c0021e761ebe6b8a42d32dfdf3dda2200f0f2fee50106c85c9aea530ccf3c3R132-R135), it seems to either return `"true"` or `"false"` (as in string, not boolean).

```
$ foo="true"
$ if [ $foo ]; then echo "is true"; else echo "is false"; fi
is true

$ foo="false"
$ if [ $foo ]; then echo "is true"; else echo "is false"; fi
is true
```

We should either change it to `if $TAG_EXISTS; then` or (more explicit and preferable IMO)  `if [[ $TAG_EXISTS == "true" ]]; then`

(nit: let's also be consistent with the use of double square brackets)

dliappis (commented):
Thanks for this work!

I didn't have any substantial comments, with the exception of one potential bug with an `if` clause.

.buildkite/scripts/build.sh
```diff
@@ -0,0 +1,20 @@
...
+check_is_arm
+
+add_bin_path
+with_go "${GOLANG_VERSION}"
+with_mage
+google_cloud_auth
+
+make -C go -f "${MAKEFILE}" build"${is_arm}"
+docker images --format "table {{.Repository}}:{{.Tag}}\t{{.Size}}" --filter=reference="${STAGING_IMAGE}/golang-crossbuild"
+docker images --format "table {{.Repository}}:{{.Tag}}\t{{.Size}}" --filter=reference="${DOCKER_REGISTRY}/beats-dev/golang-crossbuild"
```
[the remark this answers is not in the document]
oakrizan: Tried to do that, but there is no easy and fast way to do that - it will require to change almost all makefiles in the project. Taking into account that .ci will be removed quite soon, I think it's not wise to commit to that at the moment.

.buildkite/scripts/post-release.sh
oakrizan: Confirmed with  @jlind23 that force pushing is not needed anymore.

jlind23 (approved):

v1v:

PR #330, opened 2023-11-08
LLVM Apple & FPM pipelines migration

oakrizan:


.buildkite/fpm-pipeline.yml
```diff
@@ -1,3 +1,29 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: "v0.5.1"
+  GOLANG_VERSION: "1.21.3"
```
v1v, line 5: If possible, I'd say to use the `.go-version` approach so this env variable can be loaded dynamically with the `hooks` and the [golang autobump](https://github.com/elastic/golang-crossbuild/blob/main/.ci/bump-golang.yml) can be updated and simplify since it uses the GO_VERSION in the [Jenkinsfile](https://github.com/elastic/golang-crossbuild/blob/bad301b60614ba06e7ea5333357038e9ea35d396/.ci/bump-go-release-version.sh#L28).

oakrizan: Sure, thanks for the tip! Will add it.

v1v:
```suggestion
```

.buildkite/llvm-apple-pipeline.yml
```diff
@@ -1,3 +1,61 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: "v0.5.1"
+  GOLANG_VERSION: "1.21.3"
```
v1v, line 5:
```suggestion
```

jlind23 (approved):
LGTM

.buildkite/scripts/common.sh
```diff
@@ -146,3 +146,24 @@ check_is_arm() {
     is_arm=""
   fi
 }
+
+git_diff() {
+  patterns=$1
+  patterns_formatted="$(echo "$patterns" | tr ':' '\n')"
+
+  if git diff --name-only HEAD@{1} HEAD | grep -qE -e "$patterns_formatted"; then
```
dliappis, line 154: I wonder if I am missing something, but I wonder how is this supposed to work when one file doesn't match from a list in `$patterns`.

Consider the following:

```
$ git diff --name-only HEAD@{1} HEAD 
.buildkite/aarch64_pipeline.yml
.buildkite/linux_jdk_matrix_pipeline.yml
.buildkite/scripts/dra/generatesteps.py
.buildkite/scripts/jdk-matrix-tests/generate-steps.py
.buildkite/supported_plugins_test_pipeline.yml
.buildkite/windows_jdk_matrix_pipeline.yml
```

if I set `$patterns` to include a few of the above files it seems to be working as expected:

```
$ patterns=".buildkite/aarch64_pipeline.yml:.buildkite/linux_jdk_matrix_pipeline.yml"
$ patterns_formatted="$(echo "$patterns" | tr ':' '\n')"
$ git diff --name-only HEAD@{1} HEAD | grep -qE -e "$patterns_formatted"
$ echo $?
0
```

now if I introduce a deliberate typo to one of the paths, it still reports true:

```
$ patterns=".buildkite/aarch64_pipeline.yml:.buildkite/linux_jdk_matrix_pipeline.ymlTYPOHERE"
$ patterns_formatted="$(echo "$patterns" | tr ':' '\n')"
$ echo "$patterns_formatted"
.buildkite/aarch64_pipeline.yml
.buildkite/linux_jdk_matrix_pipeline.ymlTYPOHERE
$ git diff --name-only HEAD@{1} HEAD | grep -qE -e "$patterns_formatted"
$ echo $?
0
```

Is this the expected behavior?

dliappis (commented):
Generally looks good. Left a question around `git_diff()`.

.buildkite/llvm-apple-pipeline.yml
```diff
@@ -1,3 +1,60 @@
...
+    matrix:
+      setup:
+        debianVersion:
+          - "10"
+          - "11"
+          - "12"
+
+  - label: ":linux: Build LLVM Apple / Ubuntu ARM - {{matrix.debianVersion}}"
+    key: "build-ubuntu-arm"
+    if: build.env("BUILDKITE_PULL_REQUEST") != "false" || build.source == "ui" || build.branch == "main" || build.branch =~ /[0-9]+\.[0-9]+/
```
mrodm, line 40: Should it be added to the regex of branch delimiters ? 

```suggestion
    if: build.env("BUILDKITE_PULL_REQUEST") != "false" || build.source == "ui" || build.branch == "main" || build.branch =~ /^[0-9]+\.[0-9]+$$/
```

Note, `$` must be escaped. Here we use it to check for specific tags: https://github.com/elastic/elastic-package/blob/c3716bc4826521ee747fe367bd35c169e08b2080/.buildkite/pipeline.yml#L66

.buildkite/llvm-apple-pipeline.yml
```diff
@@ -1,3 +1,60 @@
...
+  MAKEFILE: "go/llvm-apple"
+  FILES_TO_CHECK: ".buildkite/llvm-apple-pipeline.yml:.buildkite/scripts/llvm-apple:go/llvm-apple"
+  DOCKER_FILTER_REF: "*/*/golang-crossbuild:llvm-apple*"
+  BUILDX: "0"
 steps:
-  - label: "Example test"
-    command: echo "Hello!"
+  - label: ":linux: Build LLVM Apple / Ubuntu X86_64 - {{matrix.debianVersion}}"
+    key: "build-ubuntu-x86"
+    if: build.env("BUILDKITE_PULL_REQUEST") != "false" || build.source == "ui" || build.branch == "main" || build.branch =~ /[0-9]+\.[0-9]+/
```
mrodm, line 17: Same as in the other step , should we add delimiters ? 
```suggestion
    if: build.env("BUILDKITE_PULL_REQUEST") != "false" || build.source == "ui" || build.branch == "main" || build.branch =~ /^[0-9]+\.[0-9]+$$/
```

.buildkite/llvm-apple-pipeline.yml
```diff
@@ -1,3 +1,60 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: "v0.5.1"
+  IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2204"
+  IMAGE_UBUNTU_ARM_64: "core-ubuntu-2004-aarch64"
+  DOCKER_REGISTRY: "docker.elastic.co"
+  STAGING_IMAGE: "${DOCKER_REGISTRY}/observability-ci"
+  DOCKER_REGISTRY_SECRET_PATH: "secret/observability-team/ci/docker-registry/prod"
```
mrodm, line 9: Is this secret available for this repository ? 

There is a secret in the ingest shared vault path:
https://github.com/elastic/package-registry/blob/d633cc4d521180f244733a8905b531eed6f75dac/.buildkite/hooks/pre-command#L12

.buildkite/llvm-apple-pipeline.yml
mrodm: There are others usages in the other pipelines, in case it is needed to be updated.

.buildkite/scripts/common.sh
```diff
@@ -146,3 +146,24 @@ check_is_arm() {
...
+}
+
+are_files_changed() {
+  patterns=$1
+
+  if [[ $(git_diff "$patterns") == false ]]; then
+    echo "No files changed in $patterns"
+    exit 0;
+  fi
+}
```
mrodm, line 168: I think it would be better that this functions returns 1 or 0 , and the caller should be the one deciding if it should exit the script.

WDYT ?

```suggestion
are_files_changed() {
  patterns=$1

  if [[ $(git_diff "$patterns") == false ]]; then
    echo "No files changed in $patterns"
    return 1
  fi
  return 0
}
```

.buildkite/scripts/llvm-fpm/publish.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source .buildkite/scripts/common.sh
+
+makefile=${1}
+patterns=${2}
+
+are_files_changed "$patterns"
```
mrodm, line 10: If this function is changed, this should also be updated

```suggestion
if ! are_files_changed "$patterns" ; then
    exit 0
fi
```

.buildkite/scripts/common.sh
oakrizan: Great point, thanks! Adjusted logic, but had to move filechange list to separate files, since [BK can't manage regexes in environment variables](https://github.com/buildkite/agent/issues/1381).  But it'd not that bad, since it's easier to read/maintain separate file.

.buildkite/hooks/pre-exit
```diff
@@ -10,7 +10,7 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" && "$BUILDKITE_STEP_KEY"
     google_cloud_logout_active_account
 fi
 
-if [[ "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" && ("$BUILDKITE_STEP_KEY" =~ ^build || "$BUILDKITE_STEP_KEY" =~ ^publish) ]]; then
+if [[ ( "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" || "$BUILDKITE_PIPELINE_SLUG" == "llvm-apple" || "$BUILDKITE_PIPELINE_SLUG" == "fpm") && ("$BUILDKITE_STEP_KEY" =~ ^build ) ]]; then
```
dliappis, line 13: Should we stay consistent with the code above i.e. either use `== build*` or `=~ ^build`?

.buildkite/fpm-pipeline.yml
```diff
@@ -1,3 +1,27 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: "v0.5.1"
+  IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2204"
+  DOCKER_REGISTRY: "docker.elastic.co"
+  STAGING_IMAGE: "${DOCKER_REGISTRY}/observability-ci"
+  MAKEFILE: "fpm"
+  FILES_TO_CHECK: ".buildkite/scripts/changeset/fpm"
```
dliappis, line 9: nit: Can this include >1 files? Or does it suport regex, hence, indeed, it may match >1 files? If yes, should we rename to `FILES_TO_CHECK_REGEX`?

dliappis (approved):
LGTM. Left two small nits.

v1v:

.buildkite/fpm-pipeline.yml
oakrizan: I have moved list of files to be checked in a separate file fpm/llvm-apple. So if any other files/directories will be necessary to check, those will be added to corresponding file. That was done, since [BK does not support regexes within the env variables](https://forum.buildkite.community/t/define-value-of-environment-variable-as-regex/1511), and fails while parsing pipeline.yml. May be the env var "FILES_TO_CHECK" name is not that accurate, I'll rename it.

mrodm (dismissed):
LGTM

oakrizan:
> There are some missing some backports, `1.20` is an active branch. Hence this PR might require the `backport-v1.20` label too.
> 
> See [#336 (comment)](https://github.com/elastic/golang-crossbuild/pull/336#issuecomment-1809892327)

@v1v - created related PR https://github.com/elastic/golang-crossbuild/pull/339

jlind23 (approved):

PR #331, opened 2023-11-08
Updated branch filtering and removed matrix

oakrizan:
1. Updated branch filtering to run on only on main & release branches
2. Removed matrix for ARM, since other Debian usage is not planned


jlind23 (approved):

v1v:

PR #335, opened 2023-11-13
updatecli: use .go-version

v1v:
### What

Use `.go-version` file to host the golang version rather than being set in the CI pipelines (Jenkinsfile or Buildkite)
Replace the updatecli to change the CI pipelines to change the `.go-version` instead

### Why

This should help with the migration to buildkite. Since the version for the Golang will be set in the `.go-version` files.

### Follow ups

Makefiles uses GO_VERSION but reading from the file it might be a bit tricky since it's relative to where the makefile is used for, hence I didn't change those Makefile but leave it as is. Updatecli changes them anyway. But Ideally it will be better if the GO_VERSION in those Makefiles could be read from the `.go-version` file instead.


.ci/bump-go-release-version.sh
```diff
@@ -24,8 +24,6 @@ GOLANG_DOWNLOAD_SHA256_ARM=$(curl -s -L https://golang.org/dl/\?mode\=json | jq
 GOLANG_DOWNLOAD_SHA256_AMD=$(curl -s -L https://golang.org/dl/\?mode\=json | jq -r ".[] | select( .version | contains(\"go${GO_RELEASE_VERSION}\")) | .files[] | select (.filename | contains(\"go${GO_RELEASE_VERSION}.linux-amd64.tar.gz\")) | .sha256")
 
 echo "Update go version ${GO_RELEASE_VERSION}"
-${SED} -E -e "s#(VERSION[[:space:]]+):= .*#\1:= ${GO_RELEASE_VERSION}#g" "go/Makefile.common"
-${SED} -E -e "s#(GO_VERSION[[:space:]]+)= .*#\1= '${GO_RELEASE_VERSION}'#g" Jenkinsfile
```
v1v, line 28: I moved this piece to the updatecli pipeline itself.

oakrizan (approved):

pierrehilbert (approved):

v1v:
@mergifyio backport 1.20

PR #352, opened 2023-11-28
Updated post-release step key

oakrizan:
Changed post release step key in pre-command script


v1v (commented):
LGTM, though, is it required a backport to `1.20`?

v1v (approved):

PR #310, opened 2023-08-14
golang: enable 1.21 

v1v:
### What

Enable 1.21 in main since it was released in early August 2023 (https://go.dev/blog/go1.21)

### Task

- Branch off `1.20` from `main`
- Use `1.21.0` golang version
- Replace `1.19` to `1.20`

### Issues

Similar to https://github.com/elastic/golang-crossbuild/pull/223

### Further details 

See [changelog](https://github.com/golang/go/issues?q=milestone%3AGo1.21+label%3ACherryPickApproved) for 1.21.0

@elastic/ingest-fp this is required when a new minor version is released as explained in https://github.com/elastic/golang-crossbuild#releasing-images-for-a-new-go-version. Can you please review this?

PR #368, opened 2023-12-10
docs: explain how to update the auto-bump

v1v:
This should help with explaining how to update the auto-bump automation, implemented in https://github.com/elastic/golang-crossbuild/pull/262,  when a new Golang version is released.

### Further details

`1.20` was created with https://github.com/elastic/golang-crossbuild/pull/310

### DIFF
--- a/.github/workflows/bump-golang-previous.yml
+++ b/.github/workflows/bump-golang-previous.yml
@@ -1,5 +1,5 @@
 ---
-name: bump-golang-1.20
+name: bump-golang-previous
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


.github/workflows/bump-golang-previous.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: bump-golang-1.20
+name: bump-golang-previous
```
v1v, line 2: version agnostic, to help with one less step to be done

amannocci (approved):

Issue #381, opened 2024-01-24
Wrong GCS bucket name in documentation


andrewkroh:
In the docs at https://github.com/elastic/golang-crossbuild/blame/43f2c4205fe38de4fecf61c39df80a9177062388/NPCAP.md#L8 the bucket name for storing npcap is documented as `gs://obs-ci-cache/private`, but the builds are looking at `gs://ingest-buildkite-ci`.

Additionally the builds are not failing when the expected binary is not found within GCS. This means that the latest set of images are broken (e.g. docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian9) because they don't have the npcap oem installer.

<img width="996" alt="Screenshot 2024-01-24 at 17 51 33" src="https://github.com/elastic/golang-crossbuild/assets/4565752/1fbbcc69-7fc3-4028-aeb2-320dcbfff171">

----

- [x] Put `gs://ingest-buildkite-ci/private/npcap-1.79-oem.exe` in place.
- [x] Overwrite the images that were published for npcap-1.79.
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian12
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian11
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian10
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian9
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79-debian8
  - [x] docker.elastic.co/observability-ci/golang-crossbuild:1.21.6-npcap-1.79
- [x] Update https://github.com/elastic/golang-crossbuild/blame/43f2c4205fe38de4fecf61c39df80a9177062388/NPCAP.md. #382

andrewkroh:
It would be awesome if there were a buildkite job to automate grabbing the latest npcap. I can provide the npcap credentials (they are in Vault).

```
# Check what the newest release is using Github.
latest_version=$(curl https://api.github.com/repos/nmap/npcap/releases/latest | jq -r .tag_name | sed 's/^v//g')

# Check if we already have the OEM installer in our GCS bucket.
if gcloud storage ls gs://ingest-buildkite-ci/private/npcap-${latest_version}-oem.exe ; then
    echo "Already present."
    exit
[...]

PR #382, opened 2024-01-25
Change GCS bucket from obs-ci-cache to ingest-buildkite-ci

andrewkroh:
Modify the documentation and Makefile.common to reflect the GCS bucket name that CI will be depending upon to hold the npcap installer exe.

Relates #381


jlind23 (approved):
Thanks for the updatE.

PR #384, opened 2024-01-25
[CI] Abort image building process if the npcap file non found

pazone:
`1.79` release caused a problem: the `npcap-1.79.exe` was missing and broken images were published despite of that. 
Now it will stop building process if the required file is missing.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -18,11 +18,11 @@ endif
 
 # Requires login at google storage.
 copy-npcap:
-	ifeq ($(CI),true)
-		@gsutil cp gs://$(GS_BUCKET_PATH)/private/$(NPCAP_FILE) ../npcap/lib/$(NPCAP_FILE) 
-	else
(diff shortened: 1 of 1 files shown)


andrewkroh (approved):
LGTM. I think there is also value in https://github.com/elastic/golang-crossbuild/pull/383 which is fail the `docker build` if the file is not present.

v1v (approved):

andrewkroh (approved):

v1v:
@pazone , this PR should be backported, too; I added the required label.

All the changes in `golang-crossbuild@main` should be backported to the branch matching the support Golang versions by default. There might be cases that are not required... but the golden-rule is backport it

Golang [supports two versions](https://go.dev/doc/devel/release#policy), that's the same principle for this particular repository.

In other words, as of today, Golang supports `1.21` and `1.20`, and that's the same in this repository, with the difference that `main` is using `1.21`.

Issue #232, opened 2022-09-09
go vcs stamping will break with new git


igor-hnizdo:
This is not an issue yet, but once you update debian to new versions, it will.

Git 2.35.2, released this April, fix a vulnerability CVE-2022-24765 ; it fixes it by requiring the user running `git` be the owner of the repo.

This affects go builds, as go run `git` to get the vcs for the vcs stamping (since go 1.18); an in this repo, the owner of the repo is different, than user inside the docker container (that is root). (Note, this is not issue on macos hosts, as they do some VM trickery.)

The fix is to use `git config --global --add safe.directory <the build folder>` as a command before the build is done. Or maybe even in the Dockerfile itself.

(This is NOT YET and issue for this repo, but might be in the future when you update debian.)

gsantoro:
This issue is now affecting my PR at https://github.com/elastic/beats/pull/34224

v1v:
https://github.com/elastic/beats/pull/34254 fixed the issue in Beats

PR #393, opened 2024-02-29
Disable VCS stamping

oakrizan:
While migrating beats repo faced an [error](https://buildkite.com/elastic/filebeat/builds/3355#018df489-70c7-4264-b0dc-c5ac70df3e1d/173-184) during packaging.
To fix issue it's required to [disable vcs stamping](https://github.com/microsoft/azurelinux/issues/3273).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,6 +7,7 @@ env:
   DOCKER_REGISTRY: "docker.elastic.co"
   STAGING_IMAGE: "${DOCKER_REGISTRY}/observability-ci"
   BUILDX: 1
+  GOFLAGS: "-buildvcs=false"
 
 steps:
   - group: "Staging"
--- a/go/base-arm/rootfs/entrypoint.go
+++ b/go/base-arm/rootfs/entrypoint.go
@@ -105,6 +105,7 @@ func buildEnvironment(platform string) (map[string]string, error) {
 		"GOARCH":      goarch,
 		"GOARM":       goarm,
 		"PLATFORM_ID": platformID,
+		"GOFLAGS": "-buildvcs=false",
 	}
 
 	if err := loadCompilerSettings(goos, arch, env); err != nil {


v1v (approved):

.buildkite/pipeline.yml
```diff
@@ -7,6 +7,7 @@ env:
   DOCKER_REGISTRY: "docker.elastic.co"
   STAGING_IMAGE: "${DOCKER_REGISTRY}/observability-ci"
   BUILDX: 1
+  GOFLAGS: "-buildvcs=false"
```
v1v, line 10: wait, IIUC, the issue is in beats if that's the case, aren't these changes needed to be included in the consumer instead?

v1v: For instance, the elastic-agent team did something similar and they are also consumers of this library (`golang-crossbuild`):
- https://github.com/elastic/elastic-agent/commit/cc81e4b376068eff7e15330f2c9fa0969d28c8f5

v1v:
In addition, https://github.com/elastic/golang-crossbuild/issues/232 might be relevant too

v1v (changes requested):
This will not fix the issue; those flags are only relevant for building the Golang Crossbuild in Buildkite. It will also not fix the problem in the consumers.

Can you point out where this is failing in the Buildkite builds for the Golang Crossbuild?

.buildkite/pipeline.yml
dliappis: > For instance, the elastic-agent team did something similar and they are also consumers of this library (`golang-crossbuild`):
> 
> * [elastic/elastic-agent@cc81e4b](https://github.com/elastic/elastic-agent/commit/cc81e4b376068eff7e15330f2c9fa0969d28c8f5)

FYI @v1v this PR you mentioned is interesting; the problem didn't get fixed by setting `export GOFLAGS='-buildvcs=false'` because this value gets overwritten and is hardcoded when docker is invoked here: https://github.com/elastic/beats/blob/main/dev-tools/mage/crossbuild.go#L330 ([permalink](https://github.com/elastic/beats/blob/e0675778b8eed65c6fff6a1d1911833c21563755/dev-tools/mage/crossbuild.go#L330)).

The problem got fixed by avoiding having a nested repo (i.e. cloning beats within another cloned repo (elastic-agent): https://github.com/elastic/elastic-agent/commit/cc81e4b376068eff7e15330f2c9fa0969d28c8f5#diff-f8bd7a8bc1f2826590a0bd7c0937871d04d15970ee75a830324888f63ce116f5R19 . A similar reproduction is described [here](https://github.com/golang/go/issues/49004#issuecomment-1172217880).

vcs stamping issues seem to be rampant after go 1.18.0, e.g. see https://github.com/golang/go/issues/49004 and specifically the commit description in  https://github.com/golang/go/commit/4569fe64101c2209e3429bd1c953b5f4021fc43d

cc @alexsapran

alexsapran:
I think we should revisit this. I am not sure why this is failing on some and working on others. 
Reading the official Golang docs

>The go command now embeds version control information in binaries. It includes the currently checked-out revision, commit time, and a flag indicating whether edited or untracked files are present. Version control information is embedded if the go command is invoked in a directory within a Git, Mercurial, Fossil, or Bazaar repository, and the main package and its containing main module are in the same repository. This information may be omitted using the flag -buildvcs=false.

So I am not 100% sure if we should by default disable it or have it in a way that allow someone to disable it, similar to how it was tried with the elastic-agent repo.

dliappis:
> I think we should revisit this. I am not sure why this is failing on some and working on others. Reading the official Golang docs
> 
> > The go command now embeds version control information in binaries. It includes the currently checked-out revision, commit time, and a flag indicating whether edited or untracked files are present. Version control information is embedded if the go command is invoked in a directory within a Git, Mercurial, Fossil, or Bazaar repository, and the main package and its containing main module are in the same repository. This information may be omitted using the flag -buildvcs=false.
> 
> So I am not 100% sure if we should by default disable it or have it in a way that allow someone to disable it, similar to how it was tried with the elastic-agent repo.

I spent a little more time during the weekend and with a [simpler reproducer](https://github.com/elastic/beats/pull/38364/files), I have able to reliable [hit the error](https://buildkite.com/elastic/filebeat/builds/3890#018e50f0-2067-49d8-9447-e36c5355398c/143-154) which [goes away](https://buildkite.com/elastic/filebeat/builds/3893#018e50fd-0f42-4d26-ad21-18b9a8230d32) when setting `-buildvcs=false` in the `GOFLAGS` of the docker image as shown in [this commit](https://github.com/elastic/beats/pull/38364/commits/d884a9c5a2515de4f5fae3155484240d2550d11f).

I also did many reruns of with buildvcs disabled and there was no flakiness.

PR #421, opened 2024-07-03
Switch to PAT generated GH token

ev1yehor:
Switch to PAT generated GH token

Linked issue:
- https://github.com/elastic/ingest-dev/issues/3495

PR #424, opened 2024-07-22
Switch to PAT generated token

ev1yehor:
Switch to PAT generated token:
Related to issue:
https://github.com/elastic/ingest-dev/issues/3495

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -6,7 +6,6 @@ source .buildkite/scripts/common.sh
 
 DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
 PRIVATE_CI_GCS_CREDENTIALS_PATH="kv/ci-shared/observability-ingest/cloud/gcp"
-GITHUB_TOKEN_VAULT_PATH="kv/ci-shared/platform-ingest/github_token"
 
 # Secrets must be redacted
 # https://buildkite.com/docs/pipelines/managing-log-output#redacted-environment-variables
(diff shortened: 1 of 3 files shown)


ev1yehor:
https://buildkite.com/elastic/golang-crossbuild/builds/421#0190fdfd-c2fc-4ace-b258-1e9e64378b4b
Was tested in this Pipeline, tag was created.

.buildkite/hooks/pre-command
```diff
@@ -22,7 +21,8 @@ if [[ ("$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" || "$BUILDKITE_PIPELINE
 fi
 
 if [[ "$BUILDKITE_PIPELINE_SLUG" == "golang-crossbuild" && "$BUILDKITE_STEP_KEY" == "release-post" ]]; then
-    export GITHUB_TOKEN_SECRET=$(retry 5 vault kv get -field token ${GITHUB_TOKEN_VAULT_PATH})
-    export GITHUB_USERNAME_SECRET=$(retry 5 vault kv get -field username ${GITHUB_TOKEN_VAULT_PATH})
-    export GITHUB_EMAIL_SECRET=$(retry 5 vault kv get -field email ${GITHUB_TOKEN_VAULT_PATH})
+    GITHUB_USERNAME_SECRET="elasticmachine"
+    export GITHUB_USERNAME_SECRET=$GITHUB_USERNAME_SECRET
+    export GITHUB_EMAIL_SECRET="elasticmachine@elastic.co"
+    export GITHUB_TOKEN_SECRET=$VAULT_GITHUB_TOKEN
```
v1v, line 27: My review in https://github.com/elastic/golang-crossbuild/pull/421 also applied here, can you please rename the variables so the `_SECRET` suffix is not needed. cc @pazone

v1v:
@ev1yehor, IIUC, the pre-command hooks are per branch no for main only, therefore the backport labes are needed here

.buildkite/hooks/pre-command
ev1yehor: Sorry that didn't answer in time. I created new PR because there was a problem to test it from fork.
I decide to not change names of variables because main issue was to change token to PAT and I didn't change names because it will follow with changes for all references to this variable in this repo. 
Also it is more secure to have SECRET in variable name which will prevent the variable from being shown in the pipeline.

ev1yehor:
@v1v my bad, I will create backports, but for which branches i should do it.

.buildkite/hooks/pre-command
v1v: > Also it is more secure to have SECRET in variable name which will prevent the variable from being shown in the pipeline.

That's true if the values are not hardcoded plainly. Buildkite pipelines are not publicly available, as far as I know, but these changes are. 

If the `_SECRET` is not honoured, it might mislead the rationale for using it in the future, for maintenance reasons, I think it's better to create a follow-up and rename those env variables.

PR #428, opened 2024-08-07
Revert "build(deps): bump updatecli/updatecli-action from 2.64.0 to 2.65.0 in /.github/actions/bump-golang in the github-actions group"

dliappis:
Reverts elastic/golang-crossbuild#427

Latest updatecli version bump causes issues (probably related to what's described in https://github.com/updatecli/updatecli/pull/2388).
This commit reverts to an older version of updatecli-action as done in other repos e.g. https://github.com/elastic/elastic-package/pull/2005

### DIFF
--- a/.github/actions/bump-golang/action.yml
+++ b/.github/actions/bump-golang/action.yml
@@ -33,7 +33,7 @@ runs:
           ref: ${{ inputs.branch }}
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v0.76.1
+        uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v2.64.0
 
       - name: Run Updatecli in Apply mode
         run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml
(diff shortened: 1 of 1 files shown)


cmacknz (dismissed):

.github/actions/bump-golang/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
           ref: ${{ inputs.branch }}
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@a0c478c868a71d3e239a65714de69450aa1ee2c6 # v0.76.1
+        uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v0.76.1
```
v1v, line 36: unrelated but if you don't mind:

```suggestion
        uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v2.64.0
```

This will fix the wrong version and help to know the real version in the upcoming bumps

dliappis: ACK I thought of this too btw. I fear that we need to update it every time updatecli raises a similar PR, since it doesn't know how to update it. Maybe better to set it to `uses: updatecli/updatecli-action@v2.64.0`?

v1v: dependabot does change it if the version is correct. In this case, the version is not matching the sha commit. My proposal will fix the issue.

For instance, see https://github.com/elastic/oblt-actions/pull/105/files

v1v (approved):

PR #429, opened 2024-08-08
updatecli: set username

v1v:
For some reason something is not right anymore

```
⚠ - 1 file(s) updated with "1.22.6":
	* go/Makefile.common
ERROR: non-fast-forward update
ERROR: push error: authentication required
```

I saw we could pass the username when interacting with [github](https://www.updatecli.io/docs/plugins/scm/github/)

image

And this solved the issue, see https://github.com/v1v/golang-crossbuild/pull/1 that verified the changes for this PR.

Maybe something has changed in the GitHub API internals?

### DIFF
--- a/.github/actions/bump-golang/action.yml
+++ b/.github/actions/bump-golang/action.yml
@@ -36,7 +36,7 @@ runs:
         uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v2.64.0
 
       - name: Run Updatecli in Apply mode
-        run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml --debug
+        run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml
         env:
           COMMAND: ${{ inputs.command }}
           BRANCH: ${{ inputs.branch }}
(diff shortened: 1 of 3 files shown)


.github/actions/bump-golang/action.yml
```diff
@@ -36,13 +36,12 @@ runs:
         uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 # v2.64.0
 
       - name: Run Updatecli in Apply mode
-        run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml
+        run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml --debug
```
dliappis, line 39: Do we need debug permanently?

v1v: I thought the same; IMO, it might help in the future, but I have no strong opinion.

dliappis: I'd suggest going the safe path, let's remove it since the output is publicly accessible and we don't know what it could potentially show (probably nothing, but as I said, let's keep safe).

v1v: GitHub secrets cannot be exposed at all, they are always masked as far as I know.
Sure

v1v:
```suggestion
        run: updatecli ${{ env.COMMAND }} --config ./.github/updatecli.d/bump-golang.yml
```

dliappis (approved):
LGTM

PR #197, opened 2022-05-12
feat: build multiarch Docker images

kuisathaverat:
it builds the base image for linux/amd64 and linux/arm64, the push is delegated to buildx, this push the manifest for the two platforms 

related to https://github.com/elastic/golang-crossbuild/issues/153

PR #436, opened 2024-08-14
Update to Go 1.23.0

v1v:
Requires https://github.com/elastic/golang-crossbuild/pull/435
Then we can branch off `main` and
Merge this PR.

Steps explained in https://github.com/elastic/golang-crossbuild#releasing-images-for-a-new-go-version

### Implementation details

- Bump version.
- Remove `install-go.sh` and use native dockerfile commands to install golang for x86 - that's the case for ARM.
- Update bump automation to avoid bumping changes in `install-go.sh`
- Update docs.

@kruskall , do you mind adding some notes regarding the changes you did to fix the below two issues:
- `can't start telemetry child process: fork/exec /usr/local/go/bin/go: invalid argument`
- `go: error obtaining buildID for go tool compile: fork/exec /usr/local/go/pkg/tool/linux_arm64/compile: invalid argument`

### DIFF
--- a/.github/updatecli.d/bump-go-release-version.sh
+++ b/.github/updatecli.d/bump-go-release-version.sh
@@ -45,9 +45,3 @@ find "go" -type f -name Dockerfile.tmpl -print0 |
             ${SED} -E -e "s#(ARG GOLANG_DOWNLOAD_SHA256)=.+#\1=${GOLANG_DOWNLOAD_SHA256_AMD}#g" "$line"
         fi
     done
-
-if [ -e go/base/install-go.sh ] ; then
-    ${SED} -E -e "s#(GOLANG_VERSION)=[0-9]+\.[0-9]+(\.[0-9]+)?#\1=${GO_RELEASE_VERSION}#g" go/base/install-go.sh
-    ${SED} -E -e "s#(GOLANG_DOWNLOAD_SHA256_AMD)=.+#\1=${GOLANG_DOWNLOAD_SHA256_AMD}#g" go/base/install-go.sh
-    ${SED} -E -e "s#(GOLANG_DOWNLOAD_SHA256_ARM)=.+#\1=${GOLANG_DOWNLOAD_SHA256_ARM}#g" go/base/install-go.sh
-fi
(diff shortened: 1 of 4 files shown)


v1v:
Thanks, @kruskall, for your contribution.

I'm unsure what's happening with the error, but it seems related to running the cross-compile for `arm` on `x86`.

go/base/Dockerfile.tmpl
```diff
@@ -31,13 +31,18 @@ RUN apt-get -o Acquire::Check-Valid-Until=false update -y --no-install-recommend
...
-  && rm /tmp/install-go.sh
+ARG GOLANG_VERSION=1.23.0
+ARG GOLANG_DOWNLOAD_URL=https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz
+ARG GOLANG_DOWNLOAD_SHA256=905a297f19ead44780548933e0ff1a1b86e8327bb459e92f9c0012569f76f5e3
+
+RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz \
+	&& echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
+        && rm -rf /usr/local/go \
+	&& tar -C /usr/local -xzf golang.tar.gz \
+	&& rm golang.tar.gz
```
v1v, line 42: What's the reason for this change? If we do so, we need to accommodate the changes in the updatecli scripts too

v1v: It was originally introduced in https://github.com/elastic/golang-crossbuild/pull/197

v1v: I've just created https://github.com/elastic/golang-crossbuild/pull/440 to update how the golang installation is done. I initially wanted to reuse the former approach with the script... but I think I might keep this proposal and remove all the references to the script.

.github/updatecli.d/bump-go-release-version.sh
```diff
@@ -45,9 +45,3 @@ find "go" -type f -name Dockerfile.tmpl -print0 |
             ${SED} -E -e "s#(ARG GOLANG_DOWNLOAD_SHA256)=.+#\1=${GOLANG_DOWNLOAD_SHA256_AMD}#g" "$line"
         fi
     done
-
-if [ -e go/base/install-go.sh ] ; then
```
v1v, line 49: we don't use this file anymore.

mauri870 (approved):

.github/updatecli.d/bump-go-release-version.sh
v1v: https://github.com/elastic/golang-crossbuild/pull/496 is the follow-up.

Issue #444, opened 2024-09-02
Cannot generate the docker images `golang 1.23` Debian 11


v1v:
```
--
  | 1.796 The following packages have unmet dependencies:
  | 1.882  libsystemd-dev:armel : Depends: libsystemd0:armel (= 247.3-7+deb11u5)
  | 1.911 E: Unable to correct problems, you have held broken packages.
  | ------

```

See https://buildkite.com/elastic/golang-crossbuild/builds/553

v1v:
https://github.com/elastic/golang-crossbuild/pull/448 is also affected

v1v:
It's blocking:
- https://github.com/elastic/golang-crossbuild/pull/443
- https://github.com/elastic/golang-crossbuild/pull/450
- https://github.com/elastic/golang-crossbuild/pull/452
- https://github.com/elastic/golang-crossbuild/pull/455

PR #451, opened 2024-09-10
Fix generating docker image

ev1yehor:
Related to issue:
https://github.com/elastic/golang-crossbuild/issues/444

### DIFF
--- a/go/armel/Dockerfile.tmpl
+++ b/go/armel/Dockerfile.tmpl
@@ -42,7 +42,7 @@ RUN apt install -y \
         librpm-dev:armel libxml2:armel
 
 # libsystemd-dev
-RUN apt-get install -y --allow-remove-essential \
+RUN apt install -y --allow-remove-essential \
         libsystemd-dev:armel libsystemd0:armel liblz4-1:armel
 {{- end }}
 
--- a/go/armel/Dockerfile.tmpl
+++ b/go/armel/Dockerfile.tmpl
@@ -42,6 +42,7 @@ RUN apt install -y \
         librpm-dev:armel libxml2:armel
 
 # libsystemd-dev
+RUN dpkg -s libsystemd0:armel | grep Version
 RUN apt install -y \
         libsystemd-dev:armel{{- if eq .DEBIAN_VERSION "11"}}=247.3-7+deb11u4{{- end }}
 {{- end }}
--- a/go/armel/Dockerfile.tmpl
+++ b/go/armel/Dockerfile.tmpl
@@ -39,7 +39,7 @@ RUN apt install -y \
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
 # librpm-dev
 RUN apt install -y \
-        librpm-dev:armel
+        librpm-dev:armel libxml2:armel
 
 # libsystemd-dev
 RUN apt install -y \
--- a/go/armel/Dockerfile.tmpl
+++ b/go/armel/Dockerfile.tmpl
@@ -43,7 +43,7 @@ RUN apt install -y \
 
 # libsystemd-dev
 RUN apt install -y \
-        libsystemd-dev:armel
+        libsystemd-dev:armel libsystemd0:armel
 {{- end }}
 
 RUN rm -rf /var/lib/apt/lists/*
--- a/go/armel/Dockerfile.tmpl
+++ b/go/armel/Dockerfile.tmpl
@@ -37,14 +37,14 @@ RUN apt install -y \
 {{- end }}
 
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
+RUN apt-cache policy libsystemd0:armel
 # librpm-dev
 RUN apt install -y \
         librpm-dev:armel libxml2:armel
 
 # libsystemd-dev
-RUN dpkg -s libsystemd0 | grep Version    
 RUN apt install -y \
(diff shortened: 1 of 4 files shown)


ev1yehor:
Which backports should I use for this PR?

v1v:
[backport-v1.22](https://github.com/elastic/golang-crossbuild/labels/backport-v1.22)


1.21 reached EoL already

v1v:
but https://github.com/elastic/golang-crossbuild/pull/443 is the one enabling the backport to 1.22, so for now, it won't work until it's merged

go/armel/Dockerfile.tmpl
```diff
@@ -39,11 +39,11 @@ RUN apt install -y \
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
 # librpm-dev
 RUN apt install -y \
-        librpm-dev:armel
+        librpm-dev:armel libxml2:armel
 
 # libsystemd-dev
-RUN apt install -y \
-        libsystemd-dev:armel
+RUN apt-get install -y --allow-remove-essential \
```
v1v, line 45: I'm not sure about this, can you elaborate what's the reason for doing so?

ev1yehor: This was done when I faced issue with upgrading package libsystemd0, you can check error in build that I will provide.
This is solution that I found now.
But Im not sure why we face this error only at debian 11. 

Example of error:
https://buildkite.com/elastic/golang-crossbuild/builds/561#0191dbad-1ea4-4a90-92ef-44558d5c86bf

kuisathaverat: It is an error and breaks the cross compilation. You will remove packages from arch x86 and replace them with armel packages. You have to find the packages of arch armel you have to install. x86 and armel packages are not compatible.

v1v: unmet dependencies is something that this project has faced a few times, once per Debian version:
- https://github.com/search?q=repo%3Aelastic%2Fgolang-crossbuild+unmet+dependencies&type=issues

Those issues might explain how things were fixed, in case it helps for fixing this issue

ev1yehor: Thank you for comments @v1v , @kuisathaverat 
I will procced to work with it now.

v1v: > IMO, these four architectures should be supported only on the latest version of Debian

I agree too

Beats use the versions below:

https://github.com/elastic/beats/blob/3a5cd592d226a5d22e9927e45a7bfeb2a98b024f/dev-tools/mage/crossbuild.go#L222-L245

I guess that's a question for the Ingest team to answer.

go/mips/Dockerfile.tmpl
```diff
@@ -46,11 +46,11 @@ RUN apt install -y \
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
 # librpm-dev
 RUN apt install -y \
-        librpm-dev:mips64el
+        librpm-dev:mips64el libxml2:mips64el
 
 # libsystemd-dev
-RUN apt install -y \
-        libsystemd-dev:mips64el
+RUN apt install -y --allow-remove-essential \
```
v1v, line 52: As discussed, this might have some side-effects

go/ppc/Dockerfile.tmpl
```diff
@@ -40,11 +40,11 @@ RUN apt install -qq -y \
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
 # librpm-dev
 RUN apt install -y \
-        librpm-dev:ppc64el
+        librpm-dev:ppc64el libxml2:ppc64el
 
 # libsystemd-dev
-RUN apt install -y \
-        libsystemd-dev:ppc64el
+RUN apt install -y --allow-remove-essential \
```
v1v, line 46: As discussed, this might have some side-effects

go/s390x/Dockerfile.tmpl
```diff
@@ -39,11 +39,11 @@ RUN apt install -qq -y \
 {{- if or (eq .DEBIAN_VERSION "10") (eq .DEBIAN_VERSION "11") (eq .DEBIAN_VERSION "12")}}
 # librpm-dev
 RUN apt install -y \
-        librpm-dev:s390x
+        librpm-dev:s390x libxml2:s390x
 
 # libsystemd-dev
-RUN apt install -y \
-        libsystemd-dev:s390x
+RUN apt install -y --allow-remove-essential \
```
v1v, line 45: As discussed, this might have some side-effects

ev1yehor:
For now I prepared PR that will disable this archs for debian 11. 
https://github.com/elastic/golang-crossbuild/pull/456
Waiting for approval that we can do it in such way and I will request review after that.

go/armel/Dockerfile.tmpl
cmacknz: Those 4 platforms (the ones removed in https://github.com/elastic/golang-crossbuild/pull/456) are not in our support matrix and we do not publish builds for them.

There may be someone out there who uses this to build their own custom projects for Debian 10 and not 12, if they do we can direct them to open a PR adding support back in a non breaking way. I don't think we have the capacity to support every permutation here for the ones that aren't actually used in Elastic products anymore.

cmacknz: +1 to removing them.

ev1yehor:
Will be solved with another PR:
https://github.com/elastic/golang-crossbuild/pull/456

PR #453, opened 2024-09-11
Test a bad bucket path will fail CI

rowlandgeoff:
Test a bad bucket path will fail CI


rowlandgeoff:
Test confirmed, closing

PR #225, opened 2022-08-18
fix: remove support for Debian 10 on armel, mips, mips32, ppc and s390x

v1v:
### What

Remove support for `Debian 10` on `armel`, `mips`, `mips32`, `ppc` and `s390x`. Recent changes on the `librpm-dev` package in `Debian 10` have broken the way we install this library for cross compile, due to we have support on `Debian 11` that works fine for those architectures (but `mips32`) we remove support for them on `Debian 10`.

### Further details 

Some packages for `mips32` are not available in `Debian 11` therefore 

image

https://packages.debian.org/bullseye/libc6-dev

vs

`Debian 10`

image

https://packages.debian.org/buster/libc6-dev

### Issues

Similar to https://github.com/elastic/golang-crossbuild/pull/175
Closes https://github.com/elastic/golang-crossbuild/issues/224

PR #175, opened 2022-04-19
fix: remove support for debian 9 on mips mips32 ppc s390x

kuisathaverat:
Remove support for Debian 9 on mips mips32 ppc s390x. Recent changes on the `librpm-dev` package in Debian 9 have broken the way we install this library for cross compile, due to we have support on Debian 10 that works fine for those architectures we remove support for them on Debian 9.

closes https://github.com/elastic/golang-crossbuild/issues/173

PR #456, opened 2024-09-13
Fix: remove support for debian 11 on armel, mips, ppc and s390x

ev1yehor:
### What

Remove support for `Debian 10` on `armel`, `mips`, `ppc` and `s390x`. Recent changes on the `libsystem-dev` package in `Debian 11` have broken the way we install this library, due to we have support on `Debian 12` that works fine for those architectures we remove support for them on `Debian 11`.

### Issues

Similar to https://github.com/elastic/golang-crossbuild/pull/175
Similar to https://github.com/elastic/golang-crossbuild/pull/225
Closes https://github.com/elastic/golang-crossbuild/issues/444

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -39,7 +39,6 @@ template: |
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-arm-debian11` - linux/armv5, linux/armv6
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-arm-debian12` - linux/armv5, linux/armv6
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-armel` - linux/armv5, linux/armv6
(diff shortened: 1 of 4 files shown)


v1v:
Please adapt the changes in the release-drafter automation:
- https://github.com/elastic/golang-crossbuild/blob/main/.github/release-drafter.yml

Otherwise, there will be references to Debian11 docker images

.github/release-drafter.yml
```diff
@@ -40,7 +40,6 @@ template: |
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-arm-debian12` - linux/armv5, linux/armv6
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-armel` - linux/armv5, linux/armv6
   - `docker.elastic.co/beats-dev/golang-crossbuild:$RESOLVED_VERSION-armel-debian9` - linux/armv5, linux/armv6
```
rowlandgeoff, line 42: To stay consistent, it seems like we should also drop armel for debian9 and update the associated README line too.

v1v:
@mergifyio backport 1.22

Issue #457, opened 2024-09-18
[golang-crossbuild] Can CI skip running when non-code related changes?


v1v:
https://github.com/elastic/golang-crossbuild/pull/450 or https://github.com/elastic/golang-crossbuild/pull/452 are changes that are not tested by the CI

@elastic/ingest-eng-prod , can the CI Buildkite pipelines be configured to run smartly what needs to be done?

rowlandgeoff:
I've done a couple test PRs and CI is still running. Will need to keep investigating.

rowlandgeoff:
I've compared the settings to the instructions at https://docs.elastic.dev/ci/configuring-pull-requests-for-buildkite and the settings in elastic-agent, where I know I've seen this functionality work. Still not sure what's going wrong.

I've set the golang-crossbuild required status check to `any source` in the branch protection rules, which I had to do in the elastic-agent repo a little while ago. 

One of the test PRs: https://github.com/elastic/golang-crossbuild/pull/475

I will also need to make the json change for the other two pipelines in pull-requests.json, since they run on PRs. But going to wait until I understand why golang-crossbuild is still running.

PR #469, opened 2024-10-04
Update pull-requests.json to skip CI in some cases

rowlandgeoff:
Skip CI when only unrelated changes are made.

Closes: https://github.com/elastic/golang-crossbuild/issues/457

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -13,7 +13,7 @@
             "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))|^/test$",
             "skip_ci_labels": [  ],
             "skip_target_branches": [ ],
-            "skip_ci_on_only_changed": [ ],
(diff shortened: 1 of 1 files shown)


v1v (commented):
Will this work with the required GH checks?

image

rowlandgeoff:
> Will this work with the required GH checks?

@v1v Good question. I branched off of this branch and opened a test PR to see if it would trigger this new flow, but it didn't. I think the pull-requests.json change must need to be merged before it can go into effect. Once this is merged, I can open a test PR soon after to verify it works.

Based on the current branch protection rules, I think I'll need to change the check source from `Buildkite` to `any source`. That's what I've seen needed in the elastic-agent repo.
image

Also, those other two required Buildkite checks can probably be removed, since they're sub-jobs of the main pipeline.

v1v (approved):
LGTM, let's see how the required checks do after merging it

alexsapran (approved):

PR #502, opened 2025-02-05
Fix segfauls and libpcap compilation errors

dliappis:
This commit fixes libpcap related compilation errors like:

```
./gencode.c:4914:13: error: 'cstate' undeclared (first use in this function)
```

by switching to a newer version of libpcap (from 1.8.1 to 1.10.4).
Additionally it resolves Segfaults in CI by using an instance type with
more memory, and finally, improves discoverability and memory usage by
switching to ubuntu 24.04 based CI images.


pkoutsovasilis (dismissed):

simitt (commented):
Left a minor note, otherwise LGTM.

simitt (dismissed):

pkoutsovasilis (approved):

v1v (dismissed):

go/armel/Dockerfile.tmpl
```diff
@@ -58,7 +58,7 @@ RUN cd / \
   && file helloWorld | cut -d "," -f 3 | grep -c 'EABI'\
   && rm helloWorld.c helloWorld
 
-RUN cd /libpcap/libpcap-1.8.1 \
+RUN cd /libpcap/libpcap-1.10.4 \
```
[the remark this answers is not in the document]
andrewkroh: Ack. I did a local build of packetbeat using go 1.23.6 with both libpcap 1.10.4 and 1.10.5 to smoke test. Packetbeat still builds and starts. I didn't do any deeper testing than that.

dliappis:
Looking at the latest run, after rebasing on `main` which includes the latest go version, I am not sure we've gotten rid of errors and segfaults that only appear on specific runs -> https://buildkite.com/elastic/golang-crossbuild/builds/748#0194d694-7fcc-462d-a401-40fba82367a3

for [example](https://buildkite.com/elastic/golang-crossbuild/builds/748#0194d689-c5e5-4497-b225-dac09a72b680/6-137):

```
38.30 configure: error: vsnprintf() is required but wasn't found
```

Seems retries (need to) kick in.

dliappis:
With this PR we are in a situation that Debian 12 consistently fails with

```
11.25 configure: error: /bin/bash ./config.sub ppc64el-unknown-linux-gnu failed
```

simitt (approved):

jlind23 (dismissed):

.buildkite/fpm-pipeline.yml
```diff
@@ -5,6 +5,7 @@ env:
   IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2204"
   DOCKER_REGISTRY: "docker.elastic.co"
   STAGING_IMAGE: "${DOCKER_REGISTRY}/observability-ci"
+  GCP_DEFAULT_X86_64_MACHINE_TYPE: "n2-standard-16"
```
dliappis, line 8: I am ambivalent whether this big jump from the default `n1-standard-4` with ~15Gi to ~64Gi -- the motivation is based on findings from interactive runs that segfauls _almost_ get eliminated already by going to 32Gi (-8).

On the cost side, based on Buildkite stats we are averaging something like 1build/week for main, so I think it's not a big problem.

jlind23 (approved):

simitt (approved):

v1v (approved):

.buildkite/fpm-pipeline.yml
simitt: If we move forward with a smaller instance and then need to trigger more rebuilds due to more failures, it might not get that much cheaper as the instance is running longer.
Maybe a smaller instance could be chosen if/when the base image gets upgrade to Ubuntu-24 as we were seeing less segfaults then.

dliappis:
Closing as it's superseded by https://github.com/elastic/golang-crossbuild/pull/505 which is a parred down version.

PR #504, opened 2025-02-06
Remove building ppc

simitt:
Building `ppc` images is not actually used and adds complexity and transient failures in the CI. The solution is to remove the `ppc` builds.


dliappis (approved):
LGTM

jlind23 (approved):

PR #505, opened 2025-02-07
Improve reliability of builds

dliappis:
This commit attempts to improve the reliability of builds from random compilation failures and segfaults by leveraging large instance types and adding a maximum of one retry upon failure.

It's essentially the same as #502 without the libpcap upgrade.

It's based on an internal discussion as a minor interim improvement until a better solution gets implemented.


jlind23 (approved):

simitt (approved):
Thanks!

Issue #509, opened 2025-02-10
Stop building debian7 and debian8


v1v:
EOL long time ago.
Delete the files in the source code
Stop building them in the CI.

Probably we could say `1.23` should not built them anymore

We have agreed to only support debian 7 and 7 in `1.22` but not anymore from `1.23.6` onwards

PR #510, opened 2025-02-10
stop publishing debian7 and debian8 from 1.23.7 onwards

v1v:
Closes https://github.com/elastic/golang-crossbuild/issues/509

See 

image


dliappis (approved):
LGTM

v1v:
@cmacknz , what do you think? IIUC, we don't use it:
- https://github.com/elastic/beats/blob/6d6067288b360df5fdb647829303f95f47ea2e10/dev-tools/mage/crossbuild.go#L222-L246. (main)
- https://github.com/elastic/beats/blob/7.17/dev-tools/mage/crossbuild.go#L196-L226 (7.17)
- https://github.com/elastic/elastic-agent/blob/45e3abf15c0fd1a351c5820646b5bd2eff6ce46b/dev-tools/mage/crossbuild.go#L223-L253 (main)

cmacknz:
You can get rid of debian8 for sure. 7.17 still references debian 7 but not 8 which is strange.

https://github.com/elastic/beats/blob/a69b50616b084595917635529e21901a44ed0171/dev-tools/mage/crossbuild.go#L196-L234

Let me look into if we still need this as the support matrix has been updated since those changes.

Main and 8.x are all using Debian 11.

cmacknz:
The Linux case selecting Debian 7 is the problematic one.

```
	case strings.HasPrefix(platform, "linux"):
		// Use an older version of libc to gain greater OS compatibility.
		// Debian 7 uses glibc 2.13.
		tagSuffix = "main-debian7"
	}
```

v1v:
Thanks for double checking, I missed the `Debian 7` in `7.17`, phew.

I'll change the scope for Debian 8 in this PR for now, and then we can figure out if `Debian7` in a follow-up

cmacknz:
From https://www.elastic.co/support/matrix

Several Beats like Packetbeat, Heartbeat, and Auditbeat claim to support Debian 8 in 7.17 with no footnotes.

For Filebeat and Metricbeat there is a footnote that we only support Debian 10 and above.

So we potentially can remove Debian 7 assuming it doesn't break compatibility with the glibc in  the older RHEL releases we support. Or we can wait until 7.17.x goes EOL and do a larger removal of lots of things.

v1v:
> So we potentially can remove Debian 7 assuming it doesn't break compatibility with the glibc in the older RHEL releases we support. Or we can wait until 7.17.x goes EOL and do a larger removal of lots of things.

I wanted to take the opportunity to remove them before `1.24` is released, so thinking the next patch for `1.23` was a good opportunity.

No rush in what we decide to do here.

We can wait for `1.24` and remove Debian 7 and Debian 8 for that Golang version - `1.24.0` is expected to be released in the upcoming weeks.

cmacknz:
We will probably leave 7.17 at 1.22 so that makes sense to me.

v1v:
> We will probably leave 7.17 at 1.22 so that makes sense to me.

Great, we have a plan, I'll merge this PR as is then

cmacknz (approved):

PR #521, opened 2025-02-26
feat(.buildkite/scripts/buildx.sh): use --progress=plain

andrewkroh:
For 'docker build buildx' add '--progress=plain' to the arguments
to prevent redrawing of the output which makes it difficult to interpret
the output when viewing CI logs.

It also disables color which makes log output easier to read.

### DIFF
--- a/.buildkite/scripts/buildx.sh
+++ b/.buildkite/scripts/buildx.sh
@@ -14,4 +14,4 @@ docker buildx create --name "${BUILDER_NAME}"
 docker buildx use "${BUILDER_NAME}"
 docker buildx inspect --bootstrap
 echo 'Build Docker image'
-docker buildx build --platform "${BUILDPLATFORM}" --push $*
+docker buildx build --progress=plain --platform "${BUILDPLATFORM}" --push $*


andrewkroh:
The failures do not appear related to the change. On the plus side, reading the errors is easier.

And I can confirm that I can now see important output such as GOARCH and GOHOSTARCH.

andrewkroh:
qemu segfault at https://buildkite.com/elastic/golang-crossbuild/builds/845#01954315-9e87-4c06-a4bf-b12d5d7a1946/112-1686

```
#10 263.1 Processing triggers for libc-bin (2.31-13+deb11u11) ...
#10 263.2 qemu: uncaught target signal 11 (Segmentation fault) - core dumped
#10 263.4 Segmentation fault (core dumped)
#10 263.4 qemu: uncaught target signal 11 (Segmentation fault) - core dumped
#10 263.5 Segmentation fault (core dumped)
#10 263.5 dpkg: error processing package libc-bin (--configure):
```

If we have amd64 and arm64 machines, then there should be no need for emulation anywhere in this pipeline, right?

dliappis:
> If we have amd64 and arm64 machines, then there should be no need for emulation anywhere in this pipeline to produce the crossbuild images, right? It might take an extra step create multi-arch images since the images are built on two separate machines.

with regard to what gets build where, looking at the [pipeline](https://github.com/elastic/golang-crossbuild/blob/main/.buildkite/pipeline.yml), or the equivalent BK visualization: https://buildkite.com/elastic/golang-crossbuild/builds/845/waterfall:

image

we can see that arm64 based workers are only used to build the debian 9.

Another puzzling element is that the command `.buildkite/scripts/build.sh|publish.sh Makefile.debian9` gets executed on both x86_64 and arm64 workers; I haven't looked into it or the Makefile, but going forwards do I understand correctly that we should find a way to:

a) not attempt to cross compile amd64/arm64? i.e. to build amd64 or arm64 variants, scripts need to run on corresponding hardware architectures
b) we shouldn't be using docker buildx on arm64, since on that particular hw architecture going forwards we should only be building arm64 images?

dliappis (approved):
LGTM -- I think we should get this in, regardless of the failures which happen anyway on main now, as it improves visibility. @v1v what do you think?

v1v (approved):
Thanks Andrew!

PR #523, opened 2025-02-26
fix(go/base): set GOTOOLCHAIN=local

andrewkroh:
Disable the automatic downloads of the Go toolchain.

For the goal of reproducible builds using golang-crossbuild, we want all of the binaries used to produce our artifacts to be self-contained in the image. Adding GOTOOLCHAIN=local disable downloads of newer toolchains.

This changes the entrypoint of the images to document that GOTOOLCHAIN is set during builds.  And it sets GOTOOLCHAIN=local in the Dockerfile so that when building entrypoint.go it uses the toolchain that was previously installed.

The two Dockerfiles (base and base-arm) had different 'go build' commands. This unifies them to both have CGO_ENABLED=0 and to explicitly set GOARCH to avoid accidentally producing a binary for the wrong target architecture (which we've seen happen for unknown reasons).

https://go.dev/doc/toolchain

PR #524, opened 2025-02-27
 fix(go/base): set GOTOOLCHAIN=local and skip building arm using the base Dockerfile

dliappis:
Disable the automatic downloads of the Go toolchain.

For the goal of reproducible builds using golang-crossbuild, we want all of the binaries used to produce our artifacts to be self-contained in the image. Adding GOTOOLCHAIN=local disable downloads of newer toolchains.

This changes the entrypoint of the images to document that GOTOOLCHAIN is set during builds.  And it sets GOTOOLCHAIN=local in the Dockerfile so that when building entrypoint.go it uses the toolchain that was previously installed.

The two Dockerfiles (base and base-arm) had different 'go build' commands. This unifies them to both have CGO_ENABLED=0 and to explicitly set GOARCH to avoid accidentally producing a binary for the wrong target architecture (which we've seen happen for unknown reasons).

Additionally we remove the arm target from the base Dockerfile, to address the CI concurrency problem described
in https://github.com/elastic/golang-crossbuild/pull/523#issuecomment-2688537270

https://go.dev/doc/toolchain

Co-Authored by @andrewkroh

### DIFF
--- a/go/Makefile.debian9
+++ b/go/Makefile.debian9
@@ -1,4 +1,4 @@
-IMAGES         := base main darwin armhf npcap
+IMAGES         := base main darwin arm armhf npcap
 ARM_IMAGES     := base-arm
 DEBIAN_VERSION := 9
 TAG_EXTENSION  := -debian9
@@ -11,7 +11,7 @@ build:
 build-arm:
 	@$(foreach var,$(ARM_IMAGES),$(MAKE) -C $(var) build-arm || exit 1;)
 
-# Requires login at https://docker-auth.elastic.co
+# Requires login at https://docker.elastic.co:7000/.
 push:
 	@$(foreach var,$(IMAGES),$(MAKE) -C $(var) push || exit 1;)
 
--- a/go/Makefile.debian9
+++ b/go/Makefile.debian9
@@ -1,4 +1,4 @@
-IMAGES         := base main darwin arm armhf npcap
+IMAGES         := base main darwin armhf npcap
 ARM_IMAGES     := base-arm
 DEBIAN_VERSION := 9
 TAG_EXTENSION  := -debian9
--- a/go/Makefile.debian9
+++ b/go/Makefile.debian9
@@ -1,4 +1,4 @@
-IMAGES         := base main darwin arm armhf npcap
+IMAGES         := base main darwin armhf npcap
 ARM_IMAGES     := base-arm
 DEBIAN_VERSION := 9
 TAG_EXTENSION  := -debian9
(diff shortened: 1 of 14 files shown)


dliappis:
To help with reviews:

this PR builds on top of #523 and makes an additional change to address the problem mentioned in https://github.com/elastic/golang-crossbuild/pull/524/files#diff-78f24cf6131fffe363db9182e23c8e5ff491034e44c49dfc5678bb85ed5146e8L1

Using this PR, `docker.elastic.co/observability-ci/golang-crossbuild:1.24.0-base-arm-debian9` gets built using the arm64 debian 9 step and doesn't get overwritten by the x86_64 debian 9 step.

The change has been tested by the image produced in the 1.23 branch (#525) and #527 ; using #527 me and @v1v tested the CI produced image `docker.elastic.co/observability-ci/golang-crossbuild:1.23.6-base-arm-debian9` in the beats PR that bumps go to 1.23.6 -> https://github.com/elastic/beats/pull/42960

go/Makefile.debian9
```diff
@@ -1,4 +1,4 @@
-IMAGES         := base main darwin arm armhf npcap
+IMAGES         := base main darwin armhf npcap
```
swiatekm, line 1: Do we want to do the same thing for `armhf` in a follow-up?

swiatekm (approved):
LGTM

v1v (approved):

go/Makefile.debian9
dliappis: Good question, I am not sure if qemu will need to be used under the hood to produce armhf binaries when running on an arm64 worker -- given that ultimately we want to avoid any involvement of qemu when running packaging jobs on arm64 workers, I felt it's safe to keep doing this on the x86_64 worker instead.

andrewkroh (approved):

v1v (dismissed):
LGTM

dliappis:
Reviewers/codeowners, I had to push a change that adds a buildkite step for the fpm image so that we have an `arm64` native version of the fpm docker image: https://github.com/elastic/golang-crossbuild/pull/524/commits/4822107b18d21af224368477ef5cc1d584fe7c61 ; with that done we could also [test elastic-agent packaging steps](https://buildkite.com/elastic/elastic-agent-extended-testing/builds/7810#_) using the arm64 staging image (`docker.elastic.co/observability-ci/fpm:1.13.1`) for fpm.

Everything else remains the same in this PR, so please take another look cc @andrewkroh / @swiatekm / @pazone

pazone (dismissed):
LGTM. Will update the [elastic-agent PR](https://github.com/elastic/elastic-agent/pull/6972) as soon as this one is merged

v1v (dismissed):

pchila (dismissed):
LGTM

jlind23 (approved):

jlind23 (approved):

PR #556, opened 2025-03-19
buildkite: support for npcap releases or urgency releases

v1v:
See https://github.com/elastic/golang-crossbuild/blob/main/NPCAP.md#backports

There is the concept of `major.minor.patch.x` for cases that need to change an existing Golang Crossbuild release with something else. 

We have used that in some cases such as the NPCAP in the past, and recently at https://github.com/elastic/golang-crossbuild/pull/554

But the existing conditions in the BK pipeline didn't allow to actually proceed.

In my view, we should allow running releases for:

- main
- ^[0-9]+.[0-9]+$
- ^[0-9]+.[0-9]+

This should help with cases such as:
- `main`
- `1.24`
- `1.24.0.x`

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -75,8 +75,8 @@ steps:
           - ".buildkite/scripts/publish.sh {{matrix.makefile}}"
         env:
           FIPS: "{{matrix.fips}}"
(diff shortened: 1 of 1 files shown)


pazone (approved):
LGTM

v1v:
@mergifyio backport 1.24.0.x

PR #565, opened 2025-03-27
Strip microsoft/go pre-release identifier

michel-laterman:
Strip static coded microsoft/go prerelease identifier from download.
The microsoft/go toolchain may include separate patches (from upstream) that are designated with a prerelease identifier, i.e., `v1.24.1-1` or `v1.24.1-2` (note: the current microsoft/go patches to `v1.24.1` only change their test infrastructure and not the binary).
Release are listed in their github: https://github.com/microsoft/go/releases
Not using a pre-release identifier uses the latest available one (this is consistent with how we gather the SHA256 checksums).

### DIFF
--- a/go/base-arm/Dockerfile.tmpl
+++ b/go/base-arm/Dockerfile.tmpl
@@ -59,7 +59,7 @@ RUN \
 
 ARG GOLANG_VERSION=1.24.1
 {{- if eq .FIPS "true"}}
-ARG GOLANG_DOWNLOAD_URL=https://aka.ms/golang/release/latest/go$GOLANG_VERSION-1.linux-arm64.tar.gz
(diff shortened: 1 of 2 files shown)


v1v (approved):
LGTM

am I right to assume we don't need to backport this to 1.24.0, right?

michel-laterman:
I don't think we need to backport to 1.24.0, our go version has been bumped across all of our repos to 1.24.1

cmacknz:
If upstream releases a patch or security fix that we do actually want, how are we detecting that? Can we setup automation to bump the version or MS patch automatically?

ycombinator (approved):

v1v:
>  Can we setup automation to bump the version or MS patch automatically?

See https://github.com/elastic/golang-crossbuild/pull/566

PR #299, opened 2023-06-15
Onboard repository in Buildkite

mrodm:
This PR onboards this repository into Buildkite adding a skeleton for the main pipeline.

PR #586, opened 2025-04-07
bk: stop cancelling any builds for the `1.*` golang branches

v1v:
https://github.com/elastic/golang-crossbuild/pull/299/ create the foundations but, `1.24` or any other `major.minor` branches should not be cancelled if a new merge commit is merged.

image

https://buildkite.com/elastic/golang-crossbuild/builds/1000#0196107a-e646-476b-aac1-a1d86ea07735


ev1yehor (approved):
LGTM

PR #588, opened 2025-04-08
updatecli: use the oblt-actions

v1v:
https://github.com/elastic/oblt-actions/tree/v1/updatecli/run-and-notify

Then we don't need PRs like https://github.com/elastic/golang-crossbuild/pull/493


pierrehilbert (approved):

ev1yehor (approved):
LGTM

PR #589, opened 2025-04-08
catalog-info: skip for changes in the .github folder

v1v:
instead of the workflows

those files are not used by BK


ev1yehor (approved):
LGTM

PR #610, opened 2025-05-08
bk: use GCP OIDC

v1v:
IIUC, BK could access the former bucket since the VM were already configured with the google credentials.

In this case, I'm changing to use OIDC for GCP and being explicit about what the BK pipelines can access to.

It requires changes in the infra part

### DIFF
--- a/NPCAP.md
+++ b/NPCAP.md
@@ -5,8 +5,8 @@ If you'd like to bump the npcap version please follow the below steps:
 1) Update `NPCAP_VERSION` value in the `Makefile`.
   * **NOTE**: Make sure the PR adding this is back-ported to the Go versions required by the Packetbeat CrossBuild target in [the mage file](https://github.com/elastic/beats/blob/main/x-pack/packetbeat/magefile.go). This is specified in the beats `.go-version` file.
 2) Download the new artifact.
-3) Upload the artifact to `gs://ingest-buildkite-ci/private`.
-  * **NOTE**: This particular Google Bucket can be accessible only by Elasticians who have got access to the Google project called `elastic-platform-ingest`.
+3) Upload the artifact to `gs://golang-crossbuild-ci-internal/private`.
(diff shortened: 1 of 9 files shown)


pkoutsovasilis:
@v1v this LGTM but since this change I believe is more infra-related I would appreciate first a review from your team 🙂

NPCAP.md
```diff
@@ -5,8 +5,8 @@ If you'd like to bump the npcap version please follow the below steps:
 1) Update `NPCAP_VERSION` value in the `Makefile`.
   * **NOTE**: Make sure the PR adding this is back-ported to the Go versions required by the Packetbeat CrossBuild target in [the mage file](https://github.com/elastic/beats/blob/main/x-pack/packetbeat/magefile.go). This is specified in the beats `.go-version` file.
 2) Download the new artifact.
-3) Upload the artifact to `gs://ingest-buildkite-ci/private`.
-  * **NOTE**: This particular Google Bucket can be accessible only by Elasticians who have got access to the Google project called `elastic-platform-ingest`.
+3) Upload the artifact to `gs://golang-crossbuild-ci-internal/private`.
```
dliappis, line 8: I think that we also need to adjust the beats repo (https://github.com/elastic/beats/blob/67636970b4225582683cca7541c27163452e59d3/x-pack/packetbeat/scripts/mage/pcap.go#L92) shortly after if we change buckets, otherwise the artifacts there e.g. `npcap` will get out of sync overtime.

v1v: See https://github.com/elastic/beats/pull/44259

dliappis (approved):
LGTM

dliappis (commented):
CI failures seem to be related to auth (`ID Token issued at <timestamp> is stale to sign-in.`) -- is this because the job is running for too long?

v1v:
> CI failures seem to be related to auth (ID Token issued at <timestamp> is stale to sign-in.) -- is this because the job is running for too long?

I've raised https://github.com/elastic/golang-crossbuild/pull/614


However, I also raised https://github.com/elastic/golang-crossbuild/issues/615, LLVM Apple pipeline was migrated, but I don't know whether it even worked after it was migrated.

PR #611, opened 2025-05-08
Revert "stop publishing debian7 and debian8 from 1.23.7 onwards (#510)"

v1v:
Notifies https://github.com/elastic/apm-server/pull/16849

So we can release `7.17` that still uses `debian-7`:
- https://github.com/elastic/beats/blob/a156701b0302ca0e4e8676d2de96ebf2c1b7879d/dev-tools/mage/crossbuild.go#L222-L225

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,6 +37,8 @@ steps:
           setup:
             makefile:
               - "Makefile"
+              - "Makefile.debian7"
+              - "Makefile.debian8"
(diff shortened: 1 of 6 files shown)


dliappis (approved):
LGTM from my side. Seems that the debian-7|8 builds were successful in CI https://buildkite.com/elastic/golang-crossbuild/builds/1067

pkoutsovasilis (approved):

PR #617, opened 2025-05-21
bk: only if targeting main

v1v:
This should avoid situations when targetting a different branch will run the FPM or llvm-apple pipelines.


dliappis (approved):
LGTM

Issue #615, opened 2025-05-20
llvm apple build image fails


v1v:
```

#13 [build_llvm_apple 7/7] RUN     cd /tmp/osxcross     && curl -sSL "https://codeload.github.com/tpoechtrager/osxcross/tar.gz/062922bbb81ac52787d8e53fa4af190acb552ec7"         \| tar -C /tmp/osxcross --strip=1 -xzf -     && patch -p1 < lcxx.patch     && UNATTENDED=1 ENABLE_COMPILER_RT_INSTALL=1 OSX_VERSION_MIN=10.10 TARGET_DIR=/usr/local/osxcross ./build.sh >/dev/null     && rm -rf "/usr/osxcross/SDK/MacOSX11.3.sdk/usr/share/man"     && tar -czf osxcross.tar.gz /usr/local/osxcross
--
  | #13 0.483 patching file wrapper/target.cpp
  | #13 ERROR: process "/bin/sh -c cd /tmp/osxcross     && curl -sSL \"https://codeload.github.com/tpoechtrager/osxcross/tar.gz/${OSXCROSS_REV}\"         \| tar -C /tmp/osxcross --strip=1 -xzf -     && patch -p1 < lcxx.patch     && UNATTENDED=1 ENABLE_COMPILER_RT_INSTALL=1 OSX_VERSION_MIN=${OSX_VERSION_MIN} TARGET_DIR=${OSXCROSS_PATH} ./build.sh >/dev/null     && rm -rf \"/usr/osxcross/SDK/${OSX_SDK}/usr/share/man\"     && tar -czf osxcross.tar.gz ${OSXCROSS_PATH}" did not complete successfully: exit code: 1
  | ------
  | > [build_llvm_apple 7/7] RUN     cd /tmp/osxcross     && curl -sSL "https://codeload.github.com/tpoechtrager/osxcross/tar.gz/062922bbb81ac52787d8e53fa4af190acb552ec7"         \| tar -C /tmp/osxcross --strip=1 -xzf -     && patch -p1 < lcxx.patch     && UNATTENDED=1 ENABLE_COMPILER_RT_INSTALL=1 OSX_VERSION_MIN=10.10 TARGET_DIR=/usr/local/osxcross ./build.sh >/dev/null     && rm -rf "/usr/osxcross/SDK/MacOSX11.3.sdk/usr/share/man"     && tar -czf osxcross.tar.gz /usr/local/osxcross:
[...]

PR #621, opened 2025-05-28
Restore debian7 images needed for beats 7.17

mauri870:
Beats 7.17 has no further planned releases, but we must remain ready for potential emergency patches.

Given this policy, we also aim to keep up with newer Go versions. PR #510 removed debian7 and debian8 images, causing failures when we upgraded 7.17 to go1.24.3. We cannot migrate to debian 9-11 due to glibc 2.13 constraints.

That said, restore the debian7 images to fix our 7.17 builds.

Relates https://github.com/elastic/beats/pull/43984.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -117,6 +117,7 @@ steps:
           setup:
             makefile:
               - "Makefile"
+              - "Makefile.debian7"
               - "Makefile.debian9"
               - "Makefile.debian10"
               - "Makefile.debian11"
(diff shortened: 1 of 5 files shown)


cmacknz:
OK, we need to keep 7.17 releasable until January 2026 which means it will will live past the EOL date of Go 1.23. We need this change but we don't need to urgently.

v1v:
unrelated test failures, see https://github.com/elastic/golang-crossbuild/issues/615

I think we need to exclude running those other pipelines for changes in the main pipeline, so I'll raise a PR to avoid that scenario.

mauri870:
@pkoutsovasilis @andrzej-stencel Could I get a review from the Control Plane team? Thanks!

pkoutsovasilis (approved):

PR #622, opened 2025-05-30
bk: avoid running llvm-apple if changes in the main pipeline.yml

v1v:
As long as https://github.com/elastic/golang-crossbuild/issues/615 , LLVM Apple build fails.

It's not an artifact we need to generate - in fact it has not been generated for > 1 year or so


mauri870 (approved):
Seems reasonable to me!

PR #625, opened 2025-06-08
validate installed golang version and honour VERSION build-args

v1v:
`VERSION` build arg are managed through https://github.com/elastic/golang-crossbuild/blob/main/go/Makefile.common#L5
Add `go version` assertion so we know if the container image has been actually produced with the correct version. See https://github.com/elastic/golang-crossbuild/pull/623#issuecomment-2951941844
Simplify the updatecli to use VERSION rather than GOLANG_VERSION.

### Why

Help with detecting breaking changes when a bump PR is partially updated. 

Still to figure out what's going on.

### DIFF
--- a/.github/updatecli.d/bump-go-release-version.sh
+++ b/.github/updatecli.d/bump-go-release-version.sh
@@ -49,7 +49,6 @@ echo "Update go version ${GO_RELEASE_VERSION}"
 
 find "go" -type f -name Dockerfile.tmpl -print0 |
     while IFS= read -r -d '' line; do
-        ${SED} -E -e "s#(ARG GOLANG_VERSION)=[0-9]+\.[0-9]+(\.[0-9]+)?#\1=${GO_RELEASE_VERSION}#g" "$line"
         if echo "$line" | grep -q 'arm' ; then
             ${SED} -E -e "s#(ARG GOLANG_DOWNLOAD_SHA256)=.+#\1=${GOLANG_DOWNLOAD_SHA256_ARM}#g" "$line"
             if [ -n "$MSFT_DOWNLOAD_SHA256_ARM" ]; then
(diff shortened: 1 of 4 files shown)


go/base/Dockerfile.tmpl
```diff
@@ -56,6 +56,10 @@ ENV PATH $PATH:/usr/local/go/bin
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 WORKDIR $GOPATH
 
+# Validate the Go installation
+RUN go version \
```
pkoutsovasilis, line 60: ok I run this locally and I validated that it works as expected 🙂

pkoutsovasilis (approved):

.github/updatecli.d/bump-go-release-version.sh
```diff
@@ -49,7 +49,6 @@ echo "Update go version ${GO_RELEASE_VERSION}"
 
 find "go" -type f -name Dockerfile.tmpl -print0 |
     while IFS= read -r -d '' line; do
-        ${SED} -E -e "s#(ARG GOLANG_VERSION)=[0-9]+\.[0-9]+(\.[0-9]+)?#\1=${GO_RELEASE_VERSION}#g" "$line"
```
pkoutsovasilis, line 52: I am not quite sure why we removed this one 🙂

go/base/Dockerfile.tmpl
v1v: Thanks for double checking, in fact, it should not fail normally, because the go installation will fail when using the `shasum` validation. But I feel it's not bad to actually add an extra step

.github/updatecli.d/bump-go-release-version.sh
pkoutsovasilis: sweet 👌

pazone (approved):

oakrizan (approved):

PR #652, opened 2025-09-04
[DRAFT] Add windows arm64 crossbuild support

dwhyrock:


nimarezainia:
> My proposal is to leave this PR open until both `beats` and `elastic-agent` project can successfully be built against this new variant (through draft PRs), and only then merge it

what exactly are those draft PRs? (i'm just curious to know when we could support arm)

PR #729, opened 2026-05-11
Fix typo in release-drafter.yml

swiatekm:
This is failing release drafts. See https://github.com/elastic/golang-crossbuild/actions/runs/25566812037.

The action was upgraded to v7, and is now less permissive about input types.

PR #730, opened 2026-05-11
Fix labels section in release-drafter.yml

michel-laterman:


michel-laterman:
Duplicate of https://github.com/elastic/golang-crossbuild/pull/729

PR #735, opened 2026-06-03
Allow running crossbuild as a non-root user

### LINKED ISSUE #732: Allow running as non-root
The crossbuild images assume the user is root inside the container. This is inconvenient when we need to handle artifacts it produces, or when mounting caches from the host. It would be preferable to run as the same UID as the host and avoid all the headaches.

As such, I'd like to adopt a similar pattern as we already have in the [fpm image](https://github.com/elastic/golang-crossbuild/blob/main/fpm/rootfs/entrypoint.sh). If the user sets particular environment variables, say `CROSSBUILD_UID` and `CROSSBUILD_GID`, then the entrypoint script sets up a temporary user with those ids and executes as that user.

The required changes are relatively small and restricted to the entrypoint, other than needing to install `su-exec` or `gosu` in the image proper. The entrypoint needs to set the likes of HOME, GOCACHE, GOMODCACHE, and that's about it.

swiatekm:
Allow the image to run with a provided UID and GID. This is done by setting the `CROSSBUILD_UID` and `CROSSBUILD_GID` environment variables. If they're set, the entrypoint creates the user and takes care of the requisite bureaucracy. If they're unset, nothing changes, other than the presence of the `su-exec` binary in the image.

Resolves https://github.com/elastic/golang-crossbuild/issues/732.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,49 @@ This will execute your projects `make build` target. While executing the build
 command the following variables with be added to the environment: GOOS, GOARCH,
 GOARM, PLATFORM_ID, CC, and CXX.
 
+### Running as a non-root user
+
+By default the container runs as root, which causes build output files on a
+bind-mounted workspace to be owned by `root` on the host. To avoid this, pass
+`CROSSBUILD_UID` and `CROSSBUILD_GID` matching your host user:
+
+```shell
+docker run -it --rm \
+  -v $GOPATH/src/github.com/user/go-project:/go/src/github.com/user/go-project \
+  -w /go/src/github.com/user/go-project \
+  -e CGO_ENABLED=1 \
+  -e CROSSBUILD_UID=$(id -u) \
+  -e CROSSBUILD_GID=$(id -g) \
+  docker.elastic.co/beats-dev/golang-crossbuild:1.16.7-armhf \
+  --build-cmd "make build" \
+  -p "linux/armv7"
+```
+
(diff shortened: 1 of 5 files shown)


v1v (commented):
LGTM.

Wonder if we could test these docker images once they are built to confirm things works as expected.

Docker images are stored in a different docker registry:
- `docker.elastic.co/observability-ci/golang-crossbuild`

and you can gather the sha256 at the very end of the build steps for each flavour

swiatekm:
> LGTM.
> 
> Wonder if we could test these docker images once they are built to confirm things works as expected.
> 
> Docker images are stored in a different docker registry:
> 
>     * `docker.elastic.co/observability-ci/golang-crossbuild`
> 
> 
> and you can gather the sha256 at the very end of the build steps for each flavour

I built the ones used in the elastic-agent build system locally and tested them there. Worked as expected. I can create a PR in that repo swapping the images to show it succeeds if you want.

v1v:
What about Beats?

swiatekm:
> What about Beats?

Opened two draft PRs switching to the new images.

v1v (approved):

samuelvl (approved):

swiatekm:
I'll wait with merging until the bests packaging steps finish in my test PR.

ycombinator:
@swiatekm @v1v should this PR be backported to the `1.26` branch?  See https://github.com/elastic/golang-crossbuild/pull/752#issuecomment-5555854433.

PR #738, opened 2026-06-26
Register Buildkite pipeline for automated npcap bump

dwhyrock:
## Summary

Registers the `golang-crossbuild-bump-npcap` Buildkite pipeline by adding it to `catalog-info.yaml`. Once merged, the RRE process will provision the pipeline at https://buildkite.com/elastic/golang-crossbuild-bump-npcap.

**This is PR 1 of 2.** A follow-up PR will add the pipeline YAML and scripts once the pipeline has been provisioned — the scripts rely on `buildkite-agent` being available, so they cannot be tested until the pipeline exists in Buildkite.

Closes https://github.com/elastic/ingest-dev/issues/2990

---

## Background

Updating the bundled npcap OEM installer in `elastic/beats` (used by Packetbeat on Windows) is currently a fully manual process:

1. A human notices a new [npcap release](https://github.com/nmap/npcap/releases)
2. They download the OEM installer from npcap.com using credentials stored in Vault
3. They upload it to `gs://golang-crossbuild-ci-internal/private/`
4. They open a PR in `elastic/golang-crossbuild` bumping `NPCAP_VERSION` in `Makefile.common` (e.g. [#710](https://github.com/elastic/golang-crossbuild/pull/710))
5. They open a PR in `elastic/beats` bumping the version in `x-pack/packetbeat/npcap/installer/LICENSE` and adding a changelog fragment (e.g. [elastic/beats#49167](https://github.com/elastic/beats/pull/49167))

This pipeline automates all five steps on a weekly schedule.

---

## What the pipeline will do (PR 2)

```
┌─────────────────────────────────────────────┐
│  Step 1: Check and upload npcap             │
│  • Query GitHub API for latest npcap tag    │
│  • Compare against NPCAP_VERSION in         │
│    Makefile.common                          │
│  • If newer: download OEM installer from    │
│    npcap.com (credentials from Vault)       │
│  • Upload to gs://golang-crossbuild-ci-     │
│    internal/private/ (GCP OIDC auth)        │
│  • Publish new version via BK meta-data     │
└───────────────────┬─────────────────────────┘
                    │ wait
          ┌─────────┴──────────┐
          ▼                    ▼
┌──────────────────┐  ┌──────────────────────┐
│ Step 2: Bump     │  │ Step 3: Bump npcap   │
│ golang-crossbuild│  │ in elastic/beats     │
│                  │  │                      │
│ • Bump           │  │ • Clone elastic/beats│
│   NPCAP_VERSION  │  │ • Bump version in    │
[...]

PR #739, opened 2026-06-30
Add automated npcap bump pipeline and scripts

dwhyrock:
## Summary

Implements the scheduled Buildkite pipeline that automates the npcap OEM installer bump, building on the pipeline registration merged in #738.

This is **PR 2 of 2**. It assumes:
- The `golang-crossbuild-bump-npcap` pipeline has been provisioned in Buildkite (via the `catalog-info.yaml` change in #738)
- NPCAP OEM credentials exist in Vault at `kv/ci-shared/observability-github-secrets/golang-crossbuild/NPCAP_USERNAME` and `NPCAP_PASSWORD` (see [elastic/observability-github-secrets#585](https://github.com/elastic/observability-github-secrets/pull/585))

Related https://github.com/elastic/ingest-dev/issues/2990

## What this adds

| File | Purpose |
|------|---------|
| `.buildkite/bump-npcap-pipeline.yml` | 2-step pipeline: upload → bump `NPCAP_VERSION` via updatecli |
| `.buildkite/scripts/bump-npcap.sh` | Queries the GitHub API for the latest npcap release, reads OEM credentials from Vault, downloads the installer (verifying it is a PE32 executable and that the MD5 matches after upload), and uploads it to GCS |
| `.buildkite/scripts/run-updatecli.sh` | Installs updatecli (version pinned in `.updatecli-version`) and applies a given config |
| `.github/updatecli.d/bump-npcap.yml` | updatecli pipeline that bumps `NPCAP_VERSION` in `Makefile.common` and opens a PR in this repo |

## How it works

1. **Check and upload** — compares the latest npcap GitHub release against `NPCAP_VERSION` in `Makefile.common`. If newer, downloads the OEM installer (HTTP digest auth via Vault-sourced credentials) and uploads it to `gs://golang-crossbuild-ci-internal/private/` using the GCP OIDC plugin. The downloaded file is verified to be a PE32 executable before upload, and the MD5 hash of the uploaded GCS object is compared against the local hash to confirm integrity. The step is idempotent: if the artifact is already in the bucket, the download/upload is skipped.
2. **Bump `NPCAP_VERSION`** runs after a `wait` step via updatecli, which independently re-derives the latest version from the GitHub releases API and opens a PR bumping `NPCAP_VERSION` in `Makefile.common`. The `elastic/vault-github-token` plugin provides the ephemeral `GITHUB_TOKEN` and configures the `elastic-ci[bot]` git identity. Because updatecli re-derives the version itself, no state needs to be passed between the two pipeline steps.
3. If no new version is found in step 1, the pipeline exits early and step 2 does not run.

## Testing

- [x] Manually trigger the pipeline via the Buildkite UI on `main` to validate the full flow end-to-end
- [x] Verify the upload step correctly skips when the GCS artifact already exists (re-run test)
- [x] Verify the downstream PR is opened with the correct version and labels
- [x] Verify the pipeline self-skips cleanly when no new version is available


.buildkite/scripts/bump-beats-npcap.sh
```diff
@@ -0,0 +1,70 @@
...
+git add "$LICENSE_FILE" "$CHANGELOG_FILE"
+git commit -m "[packetbeat] Bump npcap version to ${NPCAP_VERSION}"
+
+echo "--- Pushing branch ${BRANCH}"
+retry 3 git push \
+  "https://elasticmachine:${VAULT_GITHUB_TOKEN}@github.com/${BEATS_REPO}.git" \
+  "${BRANCH}"
+
+echo "--- Opening PR in ${BEATS_REPO}"
+GH_TOKEN="${VAULT_GITHUB_TOKEN}" gh pr create \
```
v1v: If possible, let's use https://github.com/elastic/vault-github-token-buildkite-plugin so we don't need to configure `GH_TOKEN="${VAULT_GITHUB_TOKEN}` expliciltely or setting the git config:
- https://github.com/elastic/vault-github-token-buildkite-plugin#configure-git-user-optional-boolean

.buildkite/bump-npcap-pipeline.yml
```diff
@@ -0,0 +1,50 @@
...
+  IMAGE_UBUNTU_X86_64: "family/platform-ingest-eng-prod-base-ubuntu-2204"
+  INSTANCE_TYPE_X86_64: "n2-standard-4"
+
+# This section is used to define the plugins that will be used in the pipeline.
+# See https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins
+common:
+  - oblt_google_auth_plugin: &gcp_oidc_plugin
+      # See https://github.com/elastic/oblt-infra/blob/main/conf/resources/repos/golang-crossbuild/01-gcp-buildkite-oidc.tf
+      # This plugin authenticates to Google Cloud using the OIDC token.
+      elastic/oblt-google-auth#v1.3.0:
```
v1v: https://github.com/elastic/oblt-google-auth-buildkite-plugin/releases/tag/v1.3.1

```suggestion
      elastic/oblt-google-auth#v1.3.1:
```

.buildkite/scripts/bump-npcap.sh
```diff
@@ -0,0 +1,72 @@
...
+  echo "Artifact already present: ${GCS_PATH}"
+else
+  echo "--- Downloading ${OEM_FILE} from npcap.com"
+
+  # Read OEM credentials from Vault.
+  # Path populated by: https://github.com/elastic/observability-github-secrets/pull/585
+  NPCAP_USERNAME=$(vault kv get -field=data \
+    "kv/ci-shared/observability-github-secrets/golang-crossbuild/NPCAP_USERNAME")
+  NPCAP_PASSWORD=$(vault kv get -field=data \
+    "kv/ci-shared/observability-github-secrets/golang-crossbuild/NPCAP_PASSWORD")
```
v1v: When accessing secrets directly in the commands it can expose them in the logs, to avoid so, we can use https://github.com/elastic/vault-secrets-buildkite-plugin

v1v (commented):
Wonder if using updatecli could move away from creating a PR or pushing the changes directly in the scripts:
- https://www.updatecli.io/docs/prologue/quick-start/

I somehow feel updatecli a great candidate for this case. In addition to decouple the bump in the golang-crossbuild from the beats one, so there is no need to keep the contract what's the file to be updated in Beats here.

dwhyrock:
I like the idea of decoupling this repo and beats.  I'll work on using the plugins/tools you suggested.

In the meantime, I created a PR to restore Write access for a few of us for the Beats repo so we can complete some of this automation work.  https://github.com/elastic/catalog-info/pull/4111

dwhyrock:
What permission setting do I need to update to add this pipeline to be able to access this secret?

https://buildkite.com/elastic/golang-crossbuild-bump-npcap/builds/2#019f1e55-ad87-410a-b682-e2cdd45a6f70
```
Running plugin vault-secrets-buildkite-plugin environment hook
	$ /opt/buildkite-agent/plugins/bk-agent-prod-gcp-1782920373439166229/github-com-elastic-vault-secrets-buildkite-plugin-v0-1-0/hooks/environment
	Error reading kv/data/ci-shared/observability-github-secrets/golang-crossbuild/NPCAP_USERNAME: Error making API request.
	
	URL: GET https://vault-ci-prod.elastic.dev/v1/kv/data/ci-shared/observability-github-secrets/golang-crossbuild/NPCAP_USERNAME
	Code: 403. Errors:
	
	* 1 error occurred:
		* permission denied
```

v1v:
> What permission setting do I need to update to add this pipeline to be able to access this secret?

@elastic/observablt-ci can you help Doug?

fr4nc1sc0-r4m0n:
> > What permission setting do I need to update to add this pipeline to be able to access this secret?
> 
> @elastic/observablt-ci can you help Doug?

Sure @v1v. I've been checking the permissions and everything appears to be correct:

- https://github.com/elastic/terrazzo/blob/main/manifests/prod/vault/observability-github-secrets-self-managed-golang-crossbuild.yaml

I'll keep investigating further.

**Update**: let's see if this solve the problem: https://github.com/elastic/terrazzo/pull/1424

dwhyrock:
Thanks @v1v for the help.  I tried to run it again, and it got further and was able to login and download the new npcap file.  👍 

However, when trying to upload, it didn't have permissions to the `gs://golang-crossbuild-ci-internal` bucket.

https://buildkite.com/elastic/golang-crossbuild-bump-npcap/builds/7#019f800c-587a-4af6-96c2-965f34c39dae/L202

I believe it's because this is a new pipeline with the slug `golang-crossbuild-bump-npcap`, and that slug doesn't exist in this list: https://github.com/elastic/oblt-infra/blob/main/conf/resources/repos/golang-crossbuild/01-gcp-buildkite-oidc.tf#L4-L8

Can you confirm that adding this new slug to that list will solve this permissions issue?  If so, I'll create a PR to do so.

v1v:
> Can you confirm that adding this new slug to that list will solve this permissions issue? If so, I'll create a PR to do so.

That's correct, we use OIDC to help with accessing the cloud resources, I've just raised a PR with the changes

dwhyrock:
I failed to notice that the permissions were only for read/`objectViewer`, since the new pipeline need write/`objectCreator` access to be able to upload the newly-released npcap release binary to the bucket.  

I (and Claude) created this PR to update the permissions.  Let me know if it's not correct.  https://github.com/elastic/oblt-infra/pull/898

.buildkite/scripts/bump-npcap.sh
```diff
@@ -0,0 +1,58 @@
...
+if gcloud storage ls "$GCS_PATH" 2>/dev/null; then
+  echo "Artifact already present: ${GCS_PATH}"
+else
+  echo "--- Downloading ${OEM_FILE} from npcap.com"
+  retry 3 curl -fL -O --digest -u "${NPCAP_USERNAME}:${NPCAP_PASSWORD}" \
+    "https://npcap.com/oem/dist/${OEM_FILE}"
+
+  echo "--- Uploading to ${GCS_PATH}"
+  gcloud storage cp "./${OEM_FILE}" "$GCS_PATH"
+  rm "./${OEM_FILE}"
```
efd6, line 65: What happens if the `gcloud storage cp` operation is incomplete? How does this interact with the skip-if-present check?

I don't recall if Npcap provide checksums for the artefacts, looking it seems not.

dwhyrock: Great question.  Looks like we can pass a `-no-parallel-composite-upload` flag to the cp command, which will enforce a single upload thread.  `gcloud cp` won't show any objects if the upload was canceled or interrupted, so this should be a reasonable solution.

Also unfortunate that there aren't any checksum files for the npcap downloads...

dwhyrock:
Looks like I need more access for the ephemeral github token for this pipeline.  PR created here: https://github.com/elastic/terrazzo/pull/1532

PR creation error: https://buildkite.com/elastic/golang-crossbuild-bump-npcap/builds/13#019fb4b1-ee9c-4e07-87cc-edb48869824f/L250

.buildkite/scripts/bump-npcap.sh
efd6: I think at best we can compute the checksum when we get the artefact from their server and use that for verification that it has reached the gs correctly. This is essentially what I used to do when I was doing this manually. I would imagine that the installer uses normal windows signing requirements, so we should catch issues during packetbeat CI builds, though this is a long time later than it really should ideally be.

dwhyrock:
> In addition to decouple the bump in the golang-crossbuild from the beats one, so there is no need to keep the contract what's the file to be updated in Beats here.

@v1v what are your thoughts on how to handle this in the Beats repo?  Another pipeline to check if npcap was updated in main in golang-crossbuild, and if so, create a related PR in Beats?

dwhyrock:
Example of a PR created from this PR's code: https://github.com/elastic/golang-crossbuild/pull/744

ebeahan (dismissed):
LGTM.

Recommend we get re-reviews from @efd6 @v1v to make sure all open discussion is resolved.

efd6 (dismissed):
My concerns have been addressed.

go/base/sources-debian11.list
```diff
@@ -1,4 +1,3 @@
 # see https://wiki.debian.org/CrossToolchains
-deb http://deb.debian.org/debian bullseye main
-deb http://deb.debian.org/debian bullseye-updates main
-deb http://deb.debian.org/debian-security/ bullseye-security main
+deb http://archive.debian.org/debian bullseye main
```
v1v, line 2: I'm trying to solve this in another PR:
- https://github.com/elastic/golang-crossbuild/pull/754

since it reached EOL, those sources need to be updated, but i think we should move them to another PR.

.buildkite/updatecli.d/bump-npcap.yml
```diff
@@ -0,0 +1,75 @@
...
+  githubConfig:
+    kind: github
+    spec:
+      user: 'elastic-ci[bot]'
+      username: 'elastic-ci[bot]'
+      owner: elastic
+      repository: golang-crossbuild
+      token: '{{ requiredEnv "GITHUB_TOKEN" }}'
+      branch: main
+      commitusingapi: true
```
v1v, line 23: let's honour the GH env variables, so we can keep the same approach in both cases
- https://github.com/elastic/golang-crossbuild/blob/7d23a871bd8170668ae7bf24f8eaf009bf31a10d/.github/updatecli.d/bump-golang.yml#L8-L15

.buildkite/bump-npcap-pipeline.yml
```diff
@@ -0,0 +1,65 @@
...
+    plugins:
+      - *gcp_oidc_plugin
+      - *npcap_username_plugin
+      - *npcap_password_plugin
+
+  - wait: ~
+
+  - label: ":golang: Bump NPCAP_VERSION in Makefile.common"
+    key: "npcap-bump-crossbuild"
+    command: ".buildkite/scripts/run-updatecli.sh ${UPDATECLI_CONFIG}"
```
v1v, line 56: Since we use updatecli for GH actions, could we use the same mechanism? why? Because there are already tools to help with installing updatecli in the runner and we can remove the need to maintain two different implementations when using updatecli

v1v (commented):
A few thoughts about using updatecli framework in Buidlkite

.buildkite/bump-npcap-pipeline.yml
dwhyrock: Are you suggesting we migrate this whole workflow over to Github Actions?  I believe we would need additional infra work to be able to do the external npcap check, download, and GCS upload (likely running a similar script to do it).  

Also, how do you propose we do the follow-on work to update `beats` to use the new golang-crossbuild with the new npcap?

dwhyrock (commented):
Responding to PR review...

.buildkite/updatecli.d/bump-npcap.yml
v1v: if BK then, let's revert to the previous version, I wrongly assumed it will be integrated with the existing updatecli.

.buildkite/bump-npcap-pipeline.yml
v1v: I know this is an implementation detail, but I’m trying to understand the rationale here.

IIUC, this will work in GA as you pointed out, by granting access to the npcap secrets and GCS bucket, which are things we already manage through CasC.

For Beats, I’d envision an Updatecli pipeline that fetches the npcap version and then uses Updatecli to update the version. A PR would then be raised, and if it passes, that would mean golang-crossbuild already provides that particular version for the new npcap. Otherwise, that version isn’t available yet.

I’m not blocking this at all, but I’m wondering why we’re going with two different approaches for using Updatecli here, given that the former approach is already in place. Is there a specific reason for keeping these implementations different?

PR #750, opened 2026-09-03
Update to Go 1.27.0

ebeahan:
Already created the 1.26 branch.

Bump version to 1.27.0 on main. Release notes: https://go.dev/doc/devel/release#go1.27.0


cmacknz (dismissed):

ebeahan:
Re: changes in https://github.com/elastic/golang-crossbuild/pull/750/commits/43adab24ad5b602a4ae6b33575f04950b17053fd [Per Microsoft](https://devblogs.microsoft.com/go/go-1-27-0-1-microsoft-build-now-available/), Go 1.27.0-1 removes support for `GOEXPERIMENT=systemcrypto` since it's now enabled by default

README.md
```diff
@@ -223,7 +223,7 @@ FIPS           ?=
 ```
 
 When this var is set to `"true"` the [microsoft/go](https://github.com/microsoft/go) will be used instead of the regular upstream go release.
-Additionally the docker image will have the env vars `CGO_ENABLED=1` and `GOEXPERIMENT=systemcrypto` set to ensure binaries built within the crossbuild image are FIPS compliant by default.
+Additionally the docker image will have the env var `CGO_ENABLED=1` set to ensure binaries built within the crossbuild image are FIPS compliant by default.
```
ycombinator: This is slightly misleading now.  The FIPS compliance was coming from `GOEXPERIMENT=systemcrypto`.  Now it reads like it's coming from `CGO_ENABLED=1` which is inaccurate.

ebeahan: Ok - I'll try to rephrase. Reading the [MSFT announcement](https://devblogs.microsoft.com/go/go-1-27-0-1-microsoft-build-now-available/#systemcrypto-is-no-longer-a-goexperiment-setting) for Microsoft Go 1.27, it sounds like the toolchain does this automatically.

There's a bigger conversationi about if Microsoft Go is needed now at all since Go has native FIPS 140-3 support since Go 1.24. But I want to consider that out of scope for this change.

v1v:
I think `1.27.0` should be shipped too, so every release in this project matches every release from golang. 
What's the reason `1.27.0` is not shipped first?

ebeahan:
> What's the reason 1.27.0 is not shipped first?

1.27.1 is already out, so I started there instead of the `.0`. If we want the `.0` included to follow convention, no objection to the change.

README.md
ebeahan: Improved documentations in https://github.com/elastic/golang-crossbuild/pull/750/commits/962d4f2687d298a5d5bdc007fa4c68c68f430ef5

ebeahan:
Ready for another round of reviews.

ycombinator (approved):

v1v (approved):

PR #753, opened 2026-09-06
fix: point debian11 apt sources to archive.debian.org

ycombinator:
## Summary

- Debian 11 (bullseye) reached EOL in August 2024. Its security packages have been removed from `deb.debian.org`, causing `apt-get upgrade` to fail with 404 errors during Docker image builds (e.g. [build #1944](https://buildkite.com/elastic/golang-crossbuild/builds/1944), [build #1946](https://buildkite.com/elastic/golang-crossbuild/builds/1946)).
- Switches `go/base/sources-debian11.list` and `go/base-arm/sources-debian11.list` to use `archive.debian.org`, which is the standard mirror for EOL Debian releases.
- Drops the `bullseye-updates` line (updates are folded into the main archive), consistent with how `sources-debian8.list`, `sources-debian9.list`, and `sources-debian10.list` are already structured.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go/base-arm/sources-debian11.list
+++ b/go/base-arm/sources-debian11.list
@@ -1,4 +1,4 @@
 # see https://wiki.debian.org/CrossToolchains
-deb [arch=arm64] http://deb.debian.org/debian bullseye main
-deb [arch=arm64] http://deb.debian.org/debian bullseye-updates main
-deb [arch=arm64] http://deb.debian.org/debian-security/ bullseye-security main
+# bullseye is EOL; packages are archived at archive.debian.org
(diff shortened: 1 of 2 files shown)


ycombinator:
Closing for now. `archive.debian.org/debian-security` does not yet have bullseye — it only carries stretch and buster currently (buster arrived ~1 year after its EOL). Neither approach in this PR is right: pointing at the missing security archive breaks CI with 404s, while dropping it silently skips security patches. Will reopen once the archive is available.

PR #774, opened 2026-09-16
Set llvm-apple release mode on main

v1v:
## Why
llvm-apple images were still being published to `docker.elastic.co/observability-ci` even on `main`, which meant the production namespace was not getting refreshed when the pipeline ran there.

## What changed
- Set `RELEASE=true` on the `main` llvm-apple pipeline upload step.
- This preserves the existing staging behavior for PRs and makes the llvm-apple helper scripts keep the default production repository on `main`.

## Notes
- No other image paths were changed.
- The behavior is still controlled by the existing `RELEASE` checks in `.buildkite/scripts/llvm-apple/build.sh` and `.buildkite/scripts/llvm-apple/publish.sh`.


fr4nc1sc0-r4m0n (approved):
LGTM

